PR #3, opened 2018-06-06
Add Windshaft

### LINKED ISSUE #1: Breakable toy: Add and configure Windshaft
Probably using https://github.com/azavea/docker-windshaft

mattdelsordo:
Resolves #1. The table Windshaft draws from must have a column called "the_geom_webmercator" with a SRID of 3857. I got the gardens to display as points, not sure if they're meant to display as polygons instead, but I'm fairly certain that's an issue with the database more so than Windshaft.


breakable_toy/.directory
```diff
@@ -0,0 +1,6 @@
+[Dolphin]
+Timestamp=2018,6,5,14,57,7
+Version=4
+
+[Settings]
+HiddenFilesShown=true
```
KlaasH: I think this is something your file manager made.  It shouldn't be added to the repo, and would be a good candidate for adding to a global `gitignore` file (see https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

breakable_toy/docker-compose.yml
```diff
@@ -1,6 +1,32 @@
...
+      context:
+        .
+      dockerfile: Dockerfile-windshaft
+    image: quay.io/azavea/windshaft:latest
+    depends_on:
+      redis:
+        condition:
+          service_started
+      database:
+        condition: service_started
```
KlaasH: I believe `service_started` is the default condition for `depends_on`.  Sometimes it's worth explicitly specifying settings even though they're the default (for example in a situation where even though something is the default, there's reason to believe it would come as a surprise to some future developers/maintainers), but usually it's better not to.  This actually provides a good example of one reason terser can be better--it turns out the `docker-compose` maintainers rethought things and the `condition` parameter to `depends_on` was removed.  So the simpler form
```
    depends_on:
      - redis
      - database
```
turns out to be forwards-compatible whereas the more specific version is not.

Side note: I copied and modified this file from the one in Lambnik and didn't realize I was specifying an old version of the `docker-compose` config.  I should have bumped it to 3.1.

breakable_toy/Dockerfile-windshaft
```diff
@@ -0,0 +1,34 @@
...
+    libpixman-1-0 \
+    ' \
+  && apt-get update && apt-get install -y ${buildDeps} ${deps} --no-install-recommends \
+  && rm -rf /var/lib/apt/lists/* \
+  && npm install --unsafe-perm windshaft@${WINDSHAFT_VERSION} \
+  && apt-get purge -y --auto-remove ${buildDeps}
+
+EXPOSE 5000
+
+ENTRYPOINT ["node"]
```
KlaasH: Because the [azavea/docker-windshaft](https://github.com/azavea/docker-windshaft) image is publicly available, we don't actually need the Dockerfile.  Everything this file is doing has already been done to make the published image and we can just start from that.  That's what the `image: ` line in the `docker-compose.yml` accomplishes.  (Actually, the presence of a `build` section affects the meaning of `image`. Without `build`, it means "use this existing image". With `build`, `image` changes meaning to "tag the built image with this name".)

If we did want a "similar but different" image, like if we needed to install some packages that aren't in the published image, copy/paste/modifying the Dockerfile would be one option, but in many cases it works to build on the published image.  I.e. make a Dockerfile with `FROM quay.io/azavea/windshaft:0.36.0` at the top then only specify the stuff we need in addition.  That's what the Windshaft container is doing--building on the [node:0.10-slim container](https://github.com/docker-library/node/blob/master/0.10/slim/Dockerfile).

breakable_toy/docker-compose.yml
```diff
@@ -1,6 +1,32 @@
 version: '2.1'
 services:
+  windshaft:
+    build:
+      context:
+        .
+      dockerfile: Dockerfile-windshaft
```
KlaasH: As noted above, we don't need the `build`.

breakable_toy/scripts/db-connect
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker-compose exec -u postgres database psql -d postgres
```
KlaasH: :+1: Good impulse to encapsulate the common but non-obvious tasks.  I meant to send you https://github.com/azavea/architecture/blob/master/doc/arch/adr-0000-scripts-to-rule-them-all.md, but I don't think you have access to it.  I asked Ops to add it.  Since projects often end up with quite a lot of services, our pattern is to make one script (which we call `console`, though that seems more apt for some services than others) that takes the name of the particular service as an argument.

KlaasH (commented):
This looks great.  I added some thoughts below by the things that prompted them.  One other note that doesn't quite fit on any particular line: since it's way easier to combine than sort, we tend to separate services into individual subdirectories.  So the files added for this should go in a `windshaft/` directory.  That way it's super clear what server `server.js` is, plus if we need multiple Dockerfiles they can be `windshaft/Dockerfile` rather than having to append suffixes.

PR #12, opened 2018-06-07
Add windshaft

### LINKED ISSUE #1: Breakable toy: Add and configure Windshaft

mattdelsordo:
## Overview

This PR adds Windshaft to the project, which draws from the pre-existing database to serve tiles from a template URL.

## Notes

Strictly speaking, this PR won't allow Windshaft to serve tiles properly as the example data set is not properly formatted, and altering the data seems outside the scope of this branch. In order to properly format the data, run './scripts/console db' to enter the database console and create a new column in the 'pa_gardens' table called 'the_geom_webmercator' that contains the values of the pre-existing 'geom' column re-projected to have a SRID of 3857. 
~~~~

ALTER TABLE pa_gardens ADD COLUMN the_geom_webmercator geometry(POINT, 3857);
UPDATE pa_gardens SET the_geom_webmercator = ST_Transform(ST_SetSRID(geom, 4326), 3857);

~~~~

## Testing Instructions

 * Run ./server/setup to build containers. All containers are built from images, but this is necessary in order to load in data to PostGIS.
 * Run ./scripts/server to start the containers (db, redis, and windshaft). Windshaft should tell you that map tiles are now being served.
 * Visit localhost:5000 and you should receive a default CartoDB Maps API message. Enter test values in the format http://localhost:5000/database/postgres/table/pa_gardens/{z}/{x}/{y}.png. The server should then complain that the column "the_geom_webmercator" does not exist; see Notes section for details on this.

Resolves #1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,5 @@ typings/
 
 # next.js build output
 .next
+
+.directory
--- a/breakable_toy/docker-compose.yml
+++ b/breakable_toy/docker-compose.yml
@@ -1,6 +1,25 @@
-version: '2.1'
+version: '3.1'
 services:
+  windshaft:
+    image: quay.io/azavea/windshaft:latest
+    depends_on:
+      - redis
+      - database
+    # run server entrypoint script on startup
+    command: "server/server.js"
+    volumes:
+      # map this directory's server folder to
+      # /opt/windshaft/server on the container
+      - ./windshaft:/opt/windshaft/server
+    ports:
+      # serve port 5000 on the container as port 5000 on the machine
+      - "5000:5000"
   database:
     image: quay.io/azavea/postgis:2.3-postgres9.6-slim
     expose:
+      # expose port 5432 to other containers in docker-compose
       - "5432"
+  redis:
+    image: redis:3.2
+    expose:
+    - "6379"
--- a/breakable_toy/docker-compose.yml
+++ b/breakable_toy/docker-compose.yml
@@ -4,7 +4,7 @@ services:
     image: quay.io/azavea/windshaft:latest
     depends_on:
       - redis
-      - database
+      - db
     # run server entrypoint script on startup
     command: "server/server.js"
     volumes:
@@ -14,7 +14,7 @@ services:
     ports:
       # serve port 5000 on the container as port 5000 on the machine
       - "5000:5000"
-  database:
+  db:
     image: quay.io/azavea/postgis:2.3-postgres9.6-slim
     expose:
       # expose port 5432 to other containers in docker-compose
--- a/breakable_toy/scripts/console
+++ b/breakable_toy/scripts/console
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -eu
+
+
+function usage() {
+    echo -n \
+"Usage: $(basename $0) SERVICE
+
+Create interactive console for any of the services in the project.
+
+COMMAND can be any service name as specified in docker-compose.yml.
+"
+}
+
+
+function main() {
+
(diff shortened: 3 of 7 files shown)


breakable_toy/docker-compose.yml
```diff
@@ -1,6 +1,25 @@
...
+    image: quay.io/azavea/windshaft:latest
+    depends_on:
+      - redis
+      - db
+    # run server entrypoint script on startup
+    command: "server/server.js"
+    volumes:
+      # map this directory's server folder to
+      # /opt/windshaft/server on the container
+      - ${PWD}/windshaft:/opt/windshaft/server
```
KlaasH: The host path can be relative to the directory containing the docker-compose file, i.e. `./windshaft:/opt/windshaft/server`.  Doing it that way is both slightly simpler and slightly more robust, since `${PWD}` will give the wrong answer if the server is started from some other directory (e.g. running `docker-compose -f breakable_toy/docker-compose.yml up` from the parent directory, which works otherwise).

breakable_toy/scripts/console
```diff
@@ -0,0 +1,41 @@
...
+function main() {
+
+    if [ "${1}" = "windshaft" ]
+    then
+        docker-compose exec windshaft bash
+    elif [ "${1}" = "db" ]
+    then
+        docker-compose exec -u postgres db psql -d postgres
+    else
+        echo "Unknown option ${1}! Must be a service name."
```
KlaasH: Rather than a de facto whitelist created by the comparing service names, I think this would be simpler and more adaptable if it just checked for the special cases it needs to handle (i.e. `db`) then does something generic as a fallback (i.e. `docker-compose exec ${1} bash`).  docker-compose will print a fairly helpful error for an invalid service name ("ERROR: No such service") and then new services (and the existing redis service) will just work.

breakable_toy/scripts/console
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+set -eu
+
+
+function usage() {
+    echo -n \
+"Usage: $(basename $0) [COMMAND]
```
KlaasH: Square brackets usually denote an optional argument, and I read `COMMAND` to mean the command to run (i.e. allowing alternatives to just running `bash`).  So I think this should be `"Usage: $(basename $0) SERVICE`.

.gitignore
```diff
@@ -59,3 +59,5 @@ typings/
 
 # next.js build output
 .next
+
+.directory
```
KlaasH, line 63: This is fine, but since these `.directory` files are being created by your computer, rather than anything specific to the project, putting it in a global `.gitignore` file on your computer would make sense--otherwise any project you check out and view with your file manager will end up showing these as untracked files.

breakable_toy/.gitignore
```diff
@@ -0,0 +1 @@
+.directory
```
KlaasH: Having a pattern in the `.gitignore` at the root of the project should be sufficient to get a it ignored throughout the project.

KlaasH (changes requested):
Good fixes.  The docker-compose.yml is looking great, and I got a tile with dots (at http://localhost:5000/database/postgres/table/pa_gardens/8/74/96.png)

It looks like a few files (LICENSE.txt, README.txt and the icons files) snuck in while you were adding the windshaft directory.  The simplest way to deal with that is probably to remove them then squash the removal commit in with the two existing commits related to file rearrangement.

Speaking of modifying commit history, and as just discussed, this would be a good opportunity to walk through rebasing.

KlaasH (approved):

PR #13, opened 2018-06-11
Leaflet Demo

### LINKED ISSUE #2: Breakable toy: Make a static demo page with a Leaflet map
Make a static HTML page with a demo map.
- Load Leaflet from a CDN
- Configure one or more base maps (see https://leaflet-extras.github.io/leaflet-providers/preview/)
- Add a layer showing the Windshaft tiles of our custom data

mattdelsordo:
## Overview

This PR adds a static webpage that uses Leaflet to demo the tiles drawn by Windshaft.

### Demo

leaflet-demo

### Notes

From #12:
> Strictly speaking, this PR won't allow Windshaft to serve tiles properly as the example data set is not properly formatted, and altering the data seems outside the scope of this branch. In order to properly format the data, run './scripts/console db' to enter the database console and create a new column in the 'pa_gardens' table called 'the_geom_webmercator' that contains the values of the pre-existing 'geom' column re-projected to have a SRID of 3857. 
> ~~~~
>
> ALTER TABLE pa_gardens ADD COLUMN the_geom_webmercator geometry(POINT, 3857);
> UPDATE pa_gardens SET the_geom_webmercator = ST_Transform(ST_SetSRID(geom, 4326), 3857);
>
> ~~~~

## Testing Instructions

 * Run ./server/setup to build containers. All containers are built from images, but this is necessary in order to load in data to PostGIS.
 * Run ./scripts/server to start the containers (db, redis, windshaft, leaflet). Windshaft should tell you that map tiles are now being served.
 * Visit http://localhost/. You should see a map that displays the locations from the data set.

Resolves #2


breakable_toy/docker-compose.yml
```diff
@@ -23,3 +23,14 @@ services:
     image: redis:3.2
     expose:
     - "6379"
+  leaflet:
```
KlaasH: - Blank lines are fine in YAML, and putting a blank line between service definitions helps with readability.

- It's generally good practice to match the name of a service and the directory in which you put that service's files.  Removes any ambiguity and means fewer identifiers to keep track of.

- The name can be the name of the actual program (e.g. "windshaft", "nginx", "postgres") or it can be the role it fills ("tile-server", "web", "database") or something else, if it seems like the most apt indentifier.  Some projects might have two different web servers and name them by function (e.g. "public-site" and "management-site").  In this case "webpage" or "web" seem reasonable to me.  "Leaflet" as the name of this service seems a little indirect to me, since what it's actually doing is running an nginx container that serves a wab page that loads Leaflet from a CDN and uses it.

- Notwithstanding the above, actually this demo could be done with no new service at all--we could skip having the HTML file served by the nginx container and just open the HTML in the browser directly:
image

The nginx container also works fine, and obviously if the project were going to deploy with an nginx service serving static content then it would make sense to use it for development as well.  I just wanted to point out that for a static demo like this, you can get away with very little infrastructure.

KlaasH (approved):
Looks great!  Consider either renaming the new service in `docker-compose.yml` or removing it altogether, but the demo works great and I like the icons.

breakable_toy/docker-compose.yml
mattdelsordo: That's a good point about the nginx service, my original thought was to try and compartmentalize the system by only exposing Windshaft to the nginx server rather than to a localhost port. When I realized that that isn't how that works (since you have to load the tiles from client-side JS anyway) I didn't think to get rid of the service.

PR #17, opened 2018-06-12
Add ADR 1

### LINKED ISSUE #6: Pick a Node-compatible serverless deployment framework
[Serverless Framework](https://serverless.com/) seems like the leading option, but we should evaluate how well it seems like it will suit our needs relative to the other options out there.

Wikipedia has [a short list](https://en.wikipedia.org/wiki/Serverless_computing#Serverless_frameworks), there's also an "awesome-serverless" repo with [an overwhelmingly long list](https://github.com/anaibol/awesome-serverless#frameworks).

mattdelsordo:
## Overview

Added ADR for choice of serverless framework

Resolves #6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,4 +60,5 @@ typings/
 # next.js build output
 .next
 
+# added by filesystem
 .directory
--- a/docs/adr/2018-06-12_Choose_a_serverless_framework.md
+++ b/docs/adr/2018-06-12_Choose_a_serverless_framework.md
@@ -1,0 +1,43 @@
+# ADR 1: Choose serverless framework/utility
+
+## Status
+Proposed
+
+## Context
+Lambnik is currently written in Python and uses Chalice as a framework to handle development and deployment to AWS Lambda. As Chalice is a Python-specific tool, this raises issues when porting the tool to Node.js. To that end another framework must be selected.
+
+The initial factor in this decision is a tool's functional similarity to Chalice as this should ease the translation of the project structure to the new set of technologies. As it stands, Chalice is used to deploy Lambnik and configure its connection with AWS's API Gateway. Resource provisioning is handled apart from Chalice which is an aspect of the project setup that will be preserved in order to provide for more configurability of data sources in the future. Other factors involved in choosing a framework include:
+* Ability to connect to existing project resources,
+* Ability to handle binary content in requests and responses,
+* Integration into the Node ecosystem,
+* Existence of local development tools for the framework,
+* Extent of ongoing project support,
+* Developer learning curve/support,
+* Ability to automate more of the project than was handled previously.
+
+Prospective frameworks/utilities were identified based on their ability to satisfy the functionality that Chalice currently performs, and are the following:
+1. [Serverless](#serverless)
+2. [Claudia](#claudia)
+3. [Apex](#apex)
+4. [Shep](#shep)
+
+### Serverless
+Serverless is one of the most popular serverless frameworks, its key selling point being its wide language support, service provider agnosticism, and ability to handle the configuration of an entire serverless project, including resource provisioning. This popularity comes with a high degree of customization potential with plugins, as well as robust examples and documentation. The broad language and provider support brings with it a lot of non-language-native boilerplate. There is a risk that the ability to work with a wide array of technologies comes at the cost of disobeying technology-specific conventions in order to provide a single, unified interface, which could lead to developer confusion. While Serverless is able to handle binary data such as images, [it requires multiple plugins to do so](https://medium.com/nextfaze/binary-responses-with-serverless-framework-and-api-gateway-5fde91376b76).
+
+### Claudia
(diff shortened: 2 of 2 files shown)


mmcfarland:
I'm not familiar with Claudia specifically, but I do know there are some unpleasant hacks required to get binary responses (images, in our case) out of Lambda and through the API gateway based on initial versions of the systems, for some "serverless" tools.  It would be good to confirm explicitly that this use case is supported by the selected tool (either by documentation or a proof-of-concept demo).

mattdelsordo:
That's a very good point; I did not realize that might be an issue when I was doing my research. Luckily it looks like Claudia is able to handle both binary requests and responses, according to [this article](https://claudiajs.com/tutorials/binary-content.html).

mmcfarland:
Good find.  An important callout in that article impacts our use case:

> This means that it’s currently not possible to use the API Gateway/AWS_PROXY integration to return images that can be just included into a web page using the img tag.

Since using `` is how leaflet and other map libraries place raster tile content on the map. It seems like a limitation of the platform and not the tool though, and as mentioned before, one I ran up against using Chalice.  In practice, this meant having an nginx layer in the dev environment to proxy and add appropriate `Accept` headers to requests and in "production" using CloudFront to do the same.

We may want to brace for having to do the same when developing this version.

mattdelsordo:
That is a very annoying limitation, we should definitely have a deeper chat about the implications of that at some point, assuming you think it's fine to move ahead using this tool.

adr/2018-06-12_Choose_a_serverless_framework.md
```diff
@@ -0,0 +1,42 @@
+# ADR 1: Choose serverless framework/utility
+
+## Status
+Proposed
+
+## Context
+Lambnik is currently written in Python and uses Chalice as a framework to handle development and deployment to AWS Lambda. This raises issues when porting the tool to 
+Node, however, as Chalice is a Python-specific tool. To that end another framework must be selected.
+
+The main factor influencing this decision is a tool's functional similarity to Chalice as this should ease the translation of the project structure to the new set of technologies. As it stands, Chalice is used to deploy Lambnik and configure it's connection with AWS's API Gateway. Resource provisioning is handled apart from Chalice which is an aspect of the project setup that will be preserved in order to provide for more configurability of data sources in the future. Other factors involved in choosing a framework include:
```
KlaasH: "The main factor" seems like an overstatement to me.  Seems like some of the other factors are at least as weighty.  Maybe "An initial factor"?

adr/2018-06-12_Choose_a_serverless_framework.md
```diff
@@ -0,0 +1,42 @@
...
+* Ability to connect to existing project resources
+* Integration into the Node ecosystem
+* Existence of local development tools for the framework
+* Extent of ongoing project support
+* Developer learning curve
+* Ability to automate more of the project than was handled previously
```
KlaasH: Speaking of other factors, I think it's worth including the binary handling considerations, per Matt's comments, in the list.

adr/2018-06-12_Choose_a_serverless_framework.md
KlaasH: Here and a couple places below: "it's" (contraction of "it is") -> "its" (possessive)

adr/2018-06-12_Choose_a_serverless_framework.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Decision
+We will use Claudia/Claudia API Builder as a development utility/API Framework in order to cut down on unnecessary initial bloat and learning curve steepness. Of the four options, Claudia is the closest to Chalice in terms of feature-set while operating the most natively to the Node ecosystem. While Serverless performs all the functions of Claudia and more, it also introduces boilerplate and bloat to support features that Tilegarden does not require. Additionally, compared to Serverless, Claudia requires less work to configure binary data responses, which is crucial as serving images is Tilegarden's core function. Apex and Shep, for their part, lack much of the robust documentation, workflow abstraction, and JavaScript native-ness of the other options.
+
+## Consequences
+The consequences of this decision will be:
+* Must switch frameworks if switching serverless provider
+* Potentially more tedious resource provisioning in the short term
```
KlaasH: I'm not quite clear on what this means.  Specifically "tedious" and "in the short term".  Is the idea that we'll be doing provisioning via "follow these written steps involving AWS CLI commands" initially then later automate it with a provisioning framework (e.g. terraform)?  In that case I'd say there are two pieces--additional provisioning automation work (vs. a framework that included a configurable default setup for that) and more manual provisioning in the period before we do the automation work.
Which is to say, I would replace "tedious" with something more concrete (i.e. "manual") and be clearer about what "the short term" means.

KlaasH (commented):
Looks good!  I hadn't heard of Claudia but I like the simplicity and the documentation.

Some initial notes below.

adr/2018-06-12_Choose_a_serverless_framework.md
mattdelsordo: Curses, foiled again by VS Code's lack of spell checking.

KlaasH (approved):
Looks great :+1: 
As I said over Slack, picking something besides the most popular choice might result in situations where something is harder to deal with because of the smaller feature set or developer community.  But then again it might not, and Claudia seems like a solid winner based on our current understanding of the requirements and contingencies.

PR #20, opened 2018-06-18
Feature/local dev environment

### LINKED ISSUE #8: Set up all-local development environment
If it doesn't end up getting created in the course of doing #7, add config/tooling to make a fully-local development environment.

Some priorities include:
- Ease of use setup/activation
- Fast, preferably automatic reloading
- High overlap and minimal duplication of code and config between local/development and remote/deployed

mattdelsordo:
## Overview

This PR sets up a local development environment for the project with the following features:

- Yarn to manage packages and scripts,
- Babel to transpile any ES6 code,
- ESLint for linting,
- Nodemon to hot-reload code as it is written,
- Claudia as a deployment utility,
- Claudia Local API for local lambda testing,
- Runs in a Docker container along with containers that handle a PostGIS data store and an nginx proxy.

## Notes
Serving image files from a local directory is currently not functional due to HTTP header issues. 

## Testing Instructions

 * After a clone, run `/scripts/update` to populate the database and install node modules.
 * Run `./scripts/server` to run the development server, which will be accessible on port **9001**, not 3000 as listed by the output from claudia-local-api. Several example endpoints have been defined in `src/api.js`.
 * To publish, make sure you have specified valid AWS credentials and have listed the proper database credentials in a `.env` file in the root of the project. You can then run `./scripts/publish` to publish to AWS Lambda.

Resolves #8


node_lambnik/src/tiler/proper_headers
```diff
@@ -0,0 +1,14 @@
+Connection
```
KlaasH: This looks like an accidentally-added scratch file.

node_lambnik/src/tiler/event.json
```diff
@@ -0,0 +1,3 @@
+{
```
KlaasH: This isn't used anywhere, right?

node_lambnik/src/tiler/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "tilegarden",
+  "version": "1.0.0",
+  "description": "Port of azavea/lambnik to NodeJS | AWS Lambda mapnik renderer",
+  "main": "main.js",
```
KlaasH: There's no such file.  I think we can just remove this line.

node_lambnik/src/tiler/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "bin"
+  ],
+  "scripts": {
+    "deploy-new": "yarn transpile && claudia create --region us-east-1 --role lambnik-tiler-dev --api-module bin/api --name tilegarden-${USER} --set-env POSTGRES_DB=${TILEGARDEN_DB},POSTGRES_HOST=${TILEGARDEN_DB_HOST},POSTGRES_PASSWORD=${TILEGARDEN_DB_PASSWORD},POSTGRES_USER=${TILEGARDEN_DB_USER}",
+    "test": "claudia test-lambda",
+    "lint": "eslint src",
+    "transpile": "babel src -d bin",
+    "local": "yarn transpile && claudia-local-api --api-module bin/api | bunyan -o short",
+    "dev": "nodemon --ignore bin --exec yarn local",
+    "deploy": "yarn transpile && claudia update --set-env POSTGRES_DB=${TILEGARDEN_DB},POSTGRES_HOST=${TILEGARDEN_DB_HOST},POSTGRES_PASSWORD=${TILEGARDEN_DB_PASSWORD},POSTGRES_USER=${TILEGARDEN_DB_USER}"
```
KlaasH: Probably makes sense to list these commands alphabetically.

KlaasH (commented):
Some comments below.  Still working on getting it to publish.

It would also be worth adding a basic README.md.  Not worth putting lots of effort into fleshing it out when much could change, but at least a basic guide like what's in the comment for this PR.

node_lambnik/src/proxy/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM nginx:1.12
+
+COPY etc/nginx/conf.d/default.conf /etc/nginx/conf.d/default.conf
```
KlaasH: Rather than reproduce the whole directory structure here, you could just put the conf file in this directory with the `Dockerfile` and change this line to `COPY nginx.conf /etc/nginx/conf.d/default.conf`.  Works the same and seems cleaner to me.

node_lambnik/src/tiler/proper_headers
mattdelsordo: I actually put that there to keep track of what headers the working lambnik instance was sending with its images, but yeah it probably shouldn't be in the repo.

KlaasH:
image

I guess it's not obvious because there are no timestamps there, but it's looping--rebuilding as soon as it has finished building.  It seems to happen only on the first time starting the server, so I think when the `bin/` directory didn't exist initially, the filesystem monitor keeps treating it as new.  Or something like that.

This is what I get when I try to publish:
image
I'm not sure what to make of it.

node_lambnik/README.md
```diff
@@ -0,0 +1,10 @@
+# tilegarden
+
+### Usage
+ * Make sure to have `docker`, `docker-compose`, and `yarn` installed locally.
+ * After a clone, run `./scripts/update` to populate the database and install node modules.
+ * Run `./scripts/server` to run the development server, which will be accessible on port **9001**, not 3000 as listed by the output from claudia-local-api. Several example endpoints have been defined in `src/api.js`.
+ * To publish, make sure you have specified valid AWS credentials and have listed the proper database credentials in a `.env` file in the root of the project. The format for this file can be copied from `env-template`. You can then run `./scripts/publish` to publish to AWS Lambda.
```
KlaasH, line 7: I see you have the `.aws` directory mapped in the `tiler` config in `docker-compose.yml`, but I'm not clear on what credentials it's actually looking for.  Does Claudia have a default profile name that it uses, or is it looking for `AWS_PROFILE` in the environment?  If the latter, that should probably be added to the `.env` template.

mattdelsordo: ~Claudia uses whichever AWS profile you have labeled "default", I'll make that clear in the readme and set up specifying a profile from .env.~ I've changed it to use whichever profile you have set as AWS_PROFILE in the environment.

node_lambnik/src/tiler/package.json
```diff
@@ -23,8 +23,8 @@
     "bin"
   ],
   "scripts": {
-    "deploy": "yarn transpile && claudia update --set-env POSTGRES_DB=${TILEGARDEN_DB},POSTGRES_HOST=${TILEGARDEN_DB_HOST},POSTGRES_PASSWORD=${TILEGARDEN_DB_PASSWORD},POSTGRES_USER=${TILEGARDEN_DB_USER}",
-    "deploy-new": "yarn transpile && claudia create --region us-east-1 --role lambnik-tiler-dev --api-module bin/api --name tilegarden-${USER} --set-env POSTGRES_DB=${TILEGARDEN_DB},POSTGRES_HOST=${TILEGARDEN_DB_HOST},POSTGRES_PASSWORD=${TILEGARDEN_DB_PASSWORD},POSTGRES_USER=${TILEGARDEN_DB_USER}",
+    "deploy": "yarn transpile && claudia update --profile ${AWS_PROFILE} --set-env POSTGRES_DB=${TILEGARDEN_DB},POSTGRES_HOST=${TILEGARDEN_DB_HOST},POSTGRES_PASSWORD=${TILEGARDEN_DB_PASSWORD},POSTGRES_USER=${TILEGARDEN_DB_USER}",
```
KlaasH: I think `--profile ${AWS_PROFILE}` reproduces the default behavior.  It has been working for me to run `claudia` commands with the variable set but without this argument.  Despite the wisdom of ["Explicit is better than implicit"](https://www.python.org/dev/peps/pep-0020/), I'd lean toward just letting the default behavior happen here.

node_lambnik/scripts/publish
```diff
@@ -0,0 +1,34 @@
...
+    --help      Display this help text
+"
+}
+
+function main() {
+    if [ "${1}" = "--new" ]
+    then
+        docker-compose run tiler bash -c "yarn deploy-new"
+    else
+        docker-compose run tiler bash -c "yarn deploy"
```
KlaasH: I still don't understand why `./scripts/publish` just returns "success Already up to date" for me (per the screenshot earlier) but if I change these to just run yarn rather than running bash which runs yarn, i.e. `docker-compose run tiler yarn deploy-new` and `docker-compose run tiler yarn deploy`, it works.

mattdelsordo: That's super strange, I haven't encountered that behavior at all. Have you tried doing `docker-compose down` just to clear everything before you update?

node_lambnik/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
+version: '2.1'
```
KlaasH: As on the breakable toy, I think it makes sense to start out with a more up-to-date docker-compose config version.  `3.1` seems reasonable, since only very recent versions support newer minor versions.

node_lambnik/scripts/publish
KlaasH: Yeah, no change from clearing everything and building again.  I wonder if there's a bug in my version of `docker-compose`.  It seems like it's dropping the argument and just running `yarn`.  But it works fine if I run it in the container:
image

What version of docker-compose do you have?  Mine says it's 1.17.1.
Also, do the `publish` commands that work for me work for you?  If so, we might as well do without the extra layer of `bash -c`...

node_lambnik/scripts/updatedb
```diff
@@ -0,0 +1,95 @@
...
+        h) usage; exit 0;;
+        f) force_reload=true ;;
+        \?) usage; exit 1;;
+    esac
+done
+
+check_database;
+check_postgis_install;
+
+if table_exists "pwd_inlets"; then
```
KlaasH: This is producing false positives for me, so it's skipping the import.  I believe I ran into the same problem working on setting up the database for `breakable_toy`, so the [load_db.sh](https://github.com/azavea/tilegarden/blob/develop/breakable_toy/scripts/load_db.sh) script in there is a little more robust, and should be easy to adapt (maybe just changing the default tablename).

KlaasH (changes requested):
Looks good!  Lots of moving pieces and fiddly infrastructure, but this is now working for me with the modified `publish`.

A few more small notes below, then I think this will be good to go.

node_lambnik/scripts/updatedb
mattdelsordo: I'm on docker-compose 1.17.1 as well. Those publish commands do work the same, I'll get rid of the bash part.

KlaasH (approved):
Spun up without a hitch both locally and remotely.  :tada: 
Looks great :+1:

PR #21, opened 2018-06-18
Feature/add mapnik

### LINKED ISSUE #9: Get Mapnik node package installed
https://www.npmjs.com/package/mapnik

There's a chance it will be easy, but that's generally how things turn out when installing Mapnik in any environment...

mattdelsordo:
## Overview

Adds mapnik as a dependency of the project.

### Demo

```
node_lambnik git:(feature/add-mapnik) curl localhost:9001/map
{"aspect_fix_mode":0,"parameters":{},"bufferedExtent":[null,null,null,null],"extent":[1.7976931348623157e+308,1.7976931348623157e+308,-1.7976931348623157e+308,-1.7976931348623157e+308],"bufferSize":0,"height":25,"width":25,"srs":"+init=epsg:3857"}%    
```

### Notes

~Mapnik needs to be installed from an older version of Node in order for its dependencies to install properly. I've added a `.nvmrc` file that tells `nvm` to switch to Node 8.10 (what AWS Lambda uses) when in the project directory in order to solve the problem.~ After changes to the development setup, this is no longer necessary.

## Testing Instructions

 * Run `./scripts/update` and `./scripts/server` to install dependencies and start the development server.
 * Visit [localhost:9001/map](localhost:9001/map) and you should be presented with some map metadata from mapnik.

Resolves #9

### DIFF
--- a/node_lambnik/src/tiler/package.json
+++ b/node_lambnik/src/tiler/package.json
@@ -43,6 +43,7 @@
     "nodemon": "^1.17.5"
   },
   "dependencies": {
-    "claudia-api-builder": "^4.1.0"
+    "claudia-api-builder": "^4.1.0",
+    "mapnik": "^3.7.2"
   }
 }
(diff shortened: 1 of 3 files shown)


KlaasH (approved):
Nice :+1: 
Glad it wasn't as troublesome as it initially seemed.

Issue #14, opened 2018-06-11
Add tests


KlaasH:
Set up a testing framework (very likely [Jest](https://facebook.github.io/jest/)) and add a baseline set of unit tests.

PR #23, opened 2018-06-21
Feature/tiling demo

### LINKED ISSUE #10: Implement basic tiling demo
Reimplement just the image tiles part of the [Lambnik demo map](https://github.com/azavea/lambnik/blob/develop/src/demo/index.html).

### LINKED ISSUE #19: Add proxies to make images work properly from API Gateway
API Gateway has quirks around binary content that get in the way of using it to serve tiles.  The upshot is that we need to proxy requests to the tile endpoint(s) to include an `Accept: image/png` header.
See https://github.com/azavea/tilegarden/pull/17#issuecomment-396727568 and https://claudiajs.com/tutorials/binary-content.html.

For Lambnik, this took the form of an nginx container for local development and a CloudFront distribution for the deployed endpoint.

### LINKED ISSUE #31: Flesh out error handling
Currently, server errors are just returned as a string, this can be improved.

### LINKED ISSUE #32: Make '/' display usage information
For documentation and testing purposes, to help potential users.

mattdelsordo:
## Overview
This PR plugs mapnik into the database in order to generate map tiles, which then get served up through an API.

### Demo
[Example deployment (27 Jun 2018)](https://d28f32z0lbuk2f.cloudfront.net/tile/13/2385/3103.png)
image
image

### Notes
Current development environment setup doesn't necessarily obey the quirks of the real API Gateway, tiles are served from what is basically an express server. Also, the `pwd_inlets` data set is very large on a per-tile basis which may cause the lambdas to time out. Increase the value of `TIMEOUT` in `.env` if you are consistently getting un-viewable images or 502 errors.

#### Accept headers
API Gateway currently has issues serving images that can be simply displayed in-browser, so you have to manually add an `Accept:image/png` header to your lambda requests to receive proper images. This can be done either with curl (`-H "Accept:image/png"`) or in production by setting up a proxy (e.g. with CloudFront) that adds the header to all lambda requests.

## Testing Instructions
 * Run `./scripts/update` and `./scripts/server` to run the local tileserver.
 * Open `src/demo/pwd_inlets_index.html` to generate and display tiles through Leaflet.
 * Run `./scripts/publish` (optional `--new` flag if you haven't deployed the project before). Assuming your credentials are set up properly it should deploy within about a minute.
 * Change the value of TILESERVER_PATH to load files the lambdas you just deployed.

Resolves #10, resolves #19, resolves #31, resolves #32


node_lambnik/docker-compose.yml
```diff
@@ -14,8 +14,6 @@ services:
     image: quay.io/azavea/postgis:2.3-postgres9.6-slim
     expose:
       - "5432"
-    volumes:
-      - /opt/dbdata/lambnik:/var/lib/postgresql/data
```
mmcfarland: I think the reason this is normally included is to persist the database between the cycles of `docker-compose` up/down.  I'd verify that this change still works for your normal dev workflow.

KlaasH: I suggested this because we were seeing differences between our ostensibly identical dev instances. The database here is basically a test fixture, so I think making it more ephemeral makes sense. Persisting it on the host seems to me like it adds more potential gotchas than benefits.

mattdelsordo: My stance on having that volume is that it's more of a hassle to have that data persisted and then be confused when you forget about it and need freshly create the environment than it is to have to run `update` more often, since the `update` --> `server` process is already specified in the README but it might not be totally clear that the database info is being persisted locally.

KlaasH:
This currently has demo files for both `pa_gardens` and `pwd_inlets`.  I think it would make sense to pick one to get working and leave the other out of the repo for now.  My feeling is that using `pa_gardens`, since it's simpler, is preferable for the initial demo.  That would also mean we could remove the `inlet` filtering from node-lambnik, which I think makes sense since that's very specific to the original Lambnik demo and might not match where we ultimately want to go with filtering config.

node_lambnik/src/demo/pa_gardens_index.html
```diff
@@ -0,0 +1,58 @@
...
+        // add basemap
+        //L.tileLayer('http://{s}.tile.osm.org/{z}/{x}/{y}.png', { attribution: '© OpenStreetMap' }).addTo(map)
+        L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
+            maxZoom: 19,
+            attribution: '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>'
+        }).addTo(map);
+
+
+        // add garden map
+        L.tileLayer('http://localhost:9001/tile/{z}/{x}/{y}.png').addTo(map)
```
KlaasH: Worth using a variable here, like in the original Lambnik demo, to make it easy to switch between the local and remote tilesource.

node_lambnik/src/tiler/package.json
```diff
@@ -43,6 +44,8 @@
     "nodemon": "^1.17.5"
   },
   "dependencies": {
+    "@mapbox/sphericalmercator": "^1.1.0",
+    "babel-runtime": "6",
```
KlaasH, line 47: Am I remembering correctly that you weren't sure if this was necessary?  Or am I confusing this with a conversation about something else?  In any case, where does this come into play?

node_lambnik/src/tiler/src/tile-interface.js
```diff
@@ -0,0 +1,65 @@
...
+ * TODO: this module could use a better name?
+ */
+
+import path from 'path'
+import fs from 'fs'
+
+import { image, grid } from './tiler'
+
+
+// TODO: flesh out how errors are handled
```
KlaasH: On Urban Apps, we recently had some discussions about TODO items and arrived at a process whereby any TODO added to the code (i.e. anything you're planning to merge, not just a note to be handled before finalizing a PR) should have a corresponding issue in the repo, with the issue number included in the TODO comment.  I think it's a good system, and that these TODO items are worth adding issues for.

KlaasH (commented):
Lots of good stuff here.  The image encoding thing (whatever it is) is certainly a hassle.  Having said before that it would make sense to merge this and then deal with that part, I'm now feeling like getting the basic demo page working, both locally and remotely, is probably still a better place to stop.  I had been thinking this was the `Accept` header issue, but that should be handled by the manually configured CloudFront proxy.

What's here looks good, but I think it makes sense to prune it down to 1 demo (i.e. only `pa_gardens`, as noted above) and get it to the point of working locally and remotely before closing this out.

node_lambnik/src/tiler/package.json
mattdelsordo: I got SphericalMercator working, turns out I was just telling it to use the wrong projection. Re: babel-runtime, I ran into an issue with getting Promises to run reliably after transpilation, so I added that at the internet's suggestion and it solved whatever issue that was.

node_lambnik/src/tiler/src/tile-interface.js
mattdelsordo: Sounds good, I'll add those in.

node_lambnik/src/tiler/src/api.js
```diff
@@ -3,40 +3,47 @@
...
+}
+
+const HTML_RESPONSE = { success: { contentType: 'text/html' } }
+
+// Create new lambda API
 const api = new APIBuilder()
 
-api.get('/', () => 'Hello, world!')
+// Add pngs to binary media types
+//api.setBinaryMediaTypes(['image/png'])
```
KlaasH: Looks like this was something that was tried but didn't prove useful.  Should just remove it.

node_lambnik/src/tiler/src/path-config.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Configure endpoint paths
+ */
+
+export const USAGE_PATH = '/'
+export const TILE_PATH = '/tile/{z}/{x}/{y}'
+export const GRID_PATH = '/grid/{z}/{x}/{y}'
+export const FAVICON_PATH = '/favicon.ico'
+export const IMAGE_TEST_PATH = '/img'
```
KlaasH, line 9: Looks like this was made a separate file so it can be used in both `api.js` and `tiler-interface.js`.  I think when you do #37 it should be moved back inline.  Either in the actual `api.get()` calls or as a config object at the top of `api.js`.  I'm currently leaning toward the former.

node_lambnik/src/tiler/src/tiler.js
```diff
@@ -0,0 +1,153 @@
...
+                fields: ['inlettype'],
+            }, (err, _grd) => {
+                if (err) fail(err)
+
+                return _grd.encode('utf', {
+                    resolution: 4,
+                })
+            })
+        }))
+        .then(result => result)
```
KlaasH, line 77: This seems odd.  Does it do something useful?

node_lambnik/src/tiler/src/tiler.js
```diff
@@ -0,0 +1,153 @@
...
+ * @param z
+ * @param x
+ * @param y
+ * @returns {Promise<any>}
+ */
+export const grid = (z, x, y) => {
+    const grd = new mapnik.Grid(TILE_WIDTH, TILE_HEIGHT)
+
+    return createMap(z, x, y)
+        .then((map) => new Promise((resolve, reject) => {
```
KlaasH, line 65: ~You're making a new promise here but not calling the `resolve` or `reject` functions.~
Actually, I see this is in the `grid` function and is addressed in PR #38, so that's fine.

node_lambnik/env-template
```diff
@@ -8,3 +8,9 @@ TILEGARDEN_DB=
 TILEGARDEN_DB_HOST=
 TILEGARDEN_DB_PASSWORD=
 TILEGARDEN_DB_USER=
+
+# Function config information
+DB_TABLE=
+GEOMETRY_FIELD=geom
+EPSG=4326
+TIMEOUT=5
```
KlaasH, line 16: Probably makes sense to namespace these.  E.g. `TILEGARDEN_DB_TABLE`, `TILEGARDEN_GEOMETRY_FIELD`, `TILEGARDEN_GEOMETRY_EPSG`.  And a more descriptive name for what timeout that's setting.

node_lambnik/scripts/test
```diff
@@ -0,0 +1,38 @@
...
+then
+    if [ "${1:-}" = "--help" ]
+    then
+        usage
+    else
+        SERVER_COMMAND=${1:-start}
+        main $SERVER_COMMAND
+    fi
+    exit
+fi
```
KlaasH: I think it might be better to leave this out of this PR and save it for issue #14.  Whether it stays or goes, I'd rather that it clean up after itself than create a `test-results` directory that sticks around.

KlaasH (approved):
Works well and looks good.  A few fixups and questions to consider, but :+1: with whatever you decide to do about them.

node_lambnik/env-template
mattdelsordo: I think I'm going to address this in the CartoCSS PR where I'm already refactoring a lot of the configuration setup.

node_lambnik/scripts/test
mattdelsordo: I'll remove it, all the scripts do are try to download the images from the tile server and put them in that directory, it's sort of vestigial from before the binary content situation was sorted out.

PR #35, opened 2018-06-27
Feature/debugging

### LINKED ISSUE #24: Figure out and document interactive debugging
We want to be able to step through the code running in development so we can triangulate and identify issues.  Figure out a usable set of tools and processes for doing that and add a section to the README explaining it.

mattdelsordo:
## Overview
Runs the development server in --inspect mode to expose it to debugging tools.

### Demo

image

#### (example with Chrome Dev Tools)
* Open Google Chrome and navigate to `about:inspect`.
* There should be an option listed as something along the lines of "Target (v8.10.0)" with the Node.js logo and a path like `file:///home/tiler/node_modules/claudia-local-api/bin/claudia-local-api`. Click "inspect", underneath.
 * A new window will open up. Type `ctrl+p` to open a search bar that lets you navigate to your source code. The transpiled code in `tiler/bin/` is what is actually being tracked by the debugger.

Resolves #24


node_lambnik/README.md
```diff
@@ -9,3 +9,10 @@
...
  * API Gateway has trouble serving images, so you need to configure some sort of proxy (e.g. using a CloudFront distribution) that sets an `Accept:image/png` header on all lambda requests.
+
+### Debugging
+ The local development server exposes a websocket to a node inspector that can be attached to your IDE of choice to step through and debug your code. Here are instructions on how to do so using Google Chrome's Dev Tools:
+  * Start the development server with `./scripts/server`.
+  * Open Google Chrome and navigate to `about:inspect`.
+  * There should be an option listed as something along the lines of "Target (v8.10.0)" with the Node.js logo and a path like `file:///home/tiler/node_modules/claudia-local-api/bin/claudia-local-api`. Click "inspect", underneath.
+ * A new window will open up. Type `ctrl+p` to open a search bar that lets you navigate to your source code. The transpiled code in `tiler/bin/` is what is actually being tracked by the debugger.
```
KlaasH: I added ` --source-maps inline` to the end of the `transpile` command in `package.json` and Chrome successfully found and used the source maps, so that the "Sources" tab now shows the original code.  Still worth noting that it's the transpiled code rather than the original that's being used, since that can be a gotcha, but could add that if your browser or IDE supports source maps it will look like the original source.

KlaasH (approved):
:+1: Looks good!
Don't forget to set the target branch back to `develop` (which you can do by clicking the "Edit" button by the PR title in the main tab. Not exactly intuitive) and rebase onto `develop` before merging.

PR #39, opened 2018-07-02
Feature/cartocss integration

### LINKED ISSUE #25: Support styling in CartoCSS
Add support for specifying styling in CartoCSS, converting to Mapnik XML on the fly (not during compilation/deployment).  The styling file should still be packaged with the code (we’ll expand to loading from other sources later).

### LINKED ISSUE #28: Investigate combining styling and datasource configuration

mattdelsordo:
## Overview

Overhauls the way map configurations are specified. Database connection and map layer definitions are defined in map-config.mml, which is then converted into an xml file that mapnik can read at transpile time. Secrets are templated in from the environment in order to keep them off of version control and separate between development/production.

### Notes

Templating in environment variables is good for keeping configurations separate between development and production environments, but at the end of the day secrets still end up in an XML file on AWS. This could be changed to happen at runtime avoid having credentials saved to a file but at the cost of performance, or the XML could be encrypted with something like `bcrypt` and then decrypted at runtime.

## Testing Instructions

 * Run the development server with `./scripts/update` and `./scripts/server`. 
 * Open `src/demo/pwd_inlets_index.html` to display an example map.
 * Open `src/tiler/src/res/style.mss` and change the `marker-fill` parameter. Wait for the server to restart and refresh the demo page to see the marker color update in your browser.

Resolves #25, resolves #28


node_lambnik/src/tiler/src/tiler.js
```diff
@@ -129,29 +105,29 @@ const encodeAsPNG = (renderedTile) => new Promise((resolve, reject) => {
...
-    const bounds = tileBounds(z, x, y)
-
-    // Create a webmercator map
+    // Create a webmercator map with specified bounds
     const map = new mapnik.Map(TILE_WIDTH, TILE_HEIGHT, PROJECTION)
     map.bufferSize = 64
-    map.extent = bounds
-    const layer = new mapnik.Layer('tile', PROJECTION)
-    layer.styles = ['point']
+    map.extent = tileBounds(z, x, y)
```
KlaasH: The projection is specified in the map config, so hard-coding it here is not desirable.  In addition to removing the PROJECTION constant, I think it makes sense to get rid of `tileBounds` (since it's only used this one time) and just put the `SphericalMercator().bbox` call inline.  And by inline, I mean down inside the `fromString` callback so it can use `result.srs` as the projection (since `tileBounds` is currently looking for the `EPSG` environment variable that no longer exists).

node_lambnik/src/tiler/src/tiler.js
```diff
@@ -129,29 +105,29 @@ const encodeAsPNG = (renderedTile) => new Promise((resolve, reject) => {
...
-    const postgis = getDatasource()
-    layer.datasource = postgis
-    map.add_layer(layer)
 
-    // Asynchronously load tiles from XML and return
-    return new Promise((resolve, reject) => {
-        map.load(path.join(__dirname, 'res/point-vector.xml'), { strict: true }, (err, result) => {
-            if (err) reject(err)
-            else resolve(result)
+    // load in mml and render to xml
```
KlaasH: This comment doesn't match the way this ended up being implemented.

node_lambnik/src/tiler/src/util/build-xml.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+import carto from 'carto'
+import path from 'path'
+
+import { readFile, writeFile } from './fs-promise'
+
+const IN_FILE = process.argv[2]
+const OUT_FILE = process.argv[3]
+const NODE_ENV = process.env.NODE_ENV === 'production'
```
KlaasH: This is setting a boolean "is production" flag based on the value of `NODE_ENV` in the environment.  It's a bit obfuscatory to give it the same name as the source variable when its type is different.

node_lambnik/src/tiler/src/util/build-xml.js
```diff
@@ -0,0 +1,63 @@
...
+console.log(`Production mode=${NODE_ENV}`)
+
+// Takes in a string and templates in the proper env variables
+const fillTemplate = (mmlString) => {
```
KlaasH: This defines its own basic and very specific templating syntax--converting bash-style `${VAR_NAME}` to values loaded from the environment matching the `VAR_NAME` but prefixed with the deployment environment.

I think this makes sense--this is simple and understandable, and using an existing template library or trying to make it work as a simple template string would require more work to get what we want.

I mainly want to note that if our needs get more complicated, the point at which we should switch to a defined template system rather than add nuances to this implementation would come soon.

Also, I think it's worth adding a bit of detail to the comment on this function along the lines of this note.

node_lambnik/src/tiler/src/util/build-xml.js
```diff
@@ -0,0 +1,63 @@
...
+    return mmlString.replace(/\$\{[a-z0-9_]+\}/gi, (match) => {
+        const varName = `${(NODE_ENV) ? 'PROD_' : 'DEV_'}${match.slice(2, match.length -1)}`
+        return `"${process.env[varName]}"`
+    })
+}
+
+// Loads properly formatted MML into an MML object
+const mmlToJSON = (fullMML) => {
```
KlaasH: This isn't converting MML to JSON, it's just converting an MML string to an MML object.  `loadMML` might be a better name.

KlaasH (changes requested):
As discussed in person, I think including the secrets in the XML file uploaded to S3 is fine, at least for now.  We could consider moving it back toward loading that part from the Lambda environment later if that seems useful, but the Lambda environment isn't inherently more secure than an encrypted private S3 bucket, and I think it's somewhat likely that having all the config in one place will turn out to be preferable than having some of it in the Lambda function's environment.

I think the file locations in this are a little murky.  The config is defined by the combination of `map-config.mml` and `src/res/style.mss`, then built into `src/res/map-config.xml`.  I think `map-config.mml` and `style.mss` should live in the same directory, and that it should probably be `src/config/`.  `map-config.xml` is a generated file, so it would make sense to me for it to go in `bin/`.  `src/res/` would then only have `img.png` in it, which I think we should remove.  (I kind of think "res" is too ambiguous a directory name to be useful, and that if we end up needing static images or styles or something we should either make a `static` directory or multiple directories by type of resource.)

node_lambnik/README.md
```diff
@@ -9,6 +9,9 @@
  	* Generate a UTF grid: `/grid/{z}/{x}/{y}`
  * To publish, make sure you have specified valid AWS credentials and have listed the proper database credentials in a `.env` file in the root of the project. The format for this file can be copied from `env-template`. You can then run `./scripts/publish` to publish to AWS Lambda.
  * API Gateway has trouble serving images, so you need to configure some sort of proxy (e.g. using a CloudFront distribution) that sets an `Accept:image/png` header on all lambda requests.
+ 
+### Configuration & Styling
+[Tilegarden uses CartoCSS to specify map styles.](https://carto.com/docs/carto-engine/cartocss/) Edit `src/tiler/src/map-config.mml` to specify your map configuration, including a reference any CartoCSS `.mss` files. [See here](https://cartocss.readthedocs.io/en/latest/mml.html) for Carto's `.mml` specification.
```
KlaasH: The path is now `src/tiler/src/config/map-config.mml`

node_lambnik/src/tiler/.gitignore
```diff
@@ -3,3 +3,6 @@ claudia.json
 # ignore the folder of transpiled es6
 /bin/
 tiles/
+
+# This is automatically generated and doesn't need to be committed
+map-config.xml
```
KlaasH: Now that this is built into the ephemeral `bin/` directory, this isn't needed.

node_lambnik/src/tiler/package.json
```diff
@@ -23,12 +23,13 @@
...
   "scripts": {
-    "deploy": "yarn transpile && claudia update --set-env POSTGRES_DB=${TILEGARDEN_DB},POSTGRES_HOST=${TILEGARDEN_DB_HOST},POSTGRES_PASSWORD=${TILEGARDEN_DB_PASSWORD},POSTGRES_USER=${TILEGARDEN_DB_USER},DB_TABLE=${DB_TABLE},GEOMETRY_FIELD=${GEOMETRY_FIELD},EPSG=${EPSG} --timeout ${TIMEOUT}",
-    "deploy-new": "yarn transpile && claudia create --region us-east-1 --role lambnik-tiler-dev --api-module bin/api --name tilegarden-dev-${USER} --set-env POSTGRES_DB=${TILEGARDEN_DB},POSTGRES_HOST=${TILEGARDEN_DB_HOST},POSTGRES_PASSWORD=${TILEGARDEN_DB_PASSWORD},POSTGRES_USER=${TILEGARDEN_DB_USER}DB_TABLE=${DB_TABLE},GEOMETRY_FIELD=${GEOMETRY_FIELD},EPSG=${EPSG}",
-    "dev": "nodemon -e js,xml,mss,json --ignore bin/ --exec yarn local",
+    "build-xml": "babel-node src/util/build-xml.js src/config/map-config.mml bin/map-config.xml",
+    "deploy": "yarn transpile && claudia update --timeout ${TIMEOUT}",
+    "deploy-new": "yarn transpile && claudia create --region us-east-1 --role lambnik-tiler-dev --api-module bin/api --name tilegarden-dev-${USER} --timeout ${TIMEOUT}",
+    "dev": "NODE_ENV=development && nodemon -e js,mss,json,mml,mss --ignore bin/ --exec yarn local",
```
KlaasH: `VAR=value && cmd` doesn't necessarily do what you want.  Without `export`, a variable assignment only applies to the command being run but doesn't get saved into the environment of subsequent commands.  I'm not sure exactly what the rules are for what gets the value and what doesn't when the assignment is ANDed like this, but I do know that the way to make sure everything (even invocations of other programs or scripts) on a line gets a variable is to set it as a prefix on the same command.  So what you have here, but without the `&&`.

Here's a little demo of the various behaviors.  Note how `BLAH=hi && ./test.sh` doesn't get the value.
image

node_lambnik/src/tiler/src/tile-interface.js
```diff
@@ -35,12 +35,7 @@ export const home = () => {
 }
 
 // img endpoint to verify binary content serving
-export const img = () => new Promise((resolve, reject) => {
-    fs.readFile(path.join(__dirname, 'res/img.png'), (err, data) => {
-        if (err) reject(err)
-        else resolve(data)
-    })
-})
+export const img = () => readFile(path.join(__dirname, 'res/img.png'))
```
KlaasH: I see this is dropped in #40, but since the image and its directory are removed in this PR, it would be a little cleaner to get rid of this in this one, too.

node_lambnik/src/tiler/src/tiler.js
```diff
@@ -129,29 +94,27 @@ const encodeAsPNG = (renderedTile) => new Promise((resolve, reject) => {
...
-
-    // Create a webmercator map
-    const map = new mapnik.Map(TILE_WIDTH, TILE_HEIGHT, PROJECTION)
+    // Create a webmercator map with specified bounds
+    const map = new mapnik.Map(TILE_WIDTH, TILE_HEIGHT)
     map.bufferSize = 64
-    map.extent = bounds
-    const layer = new mapnik.Layer('tile', PROJECTION)
-    layer.styles = ['point']
+    map.extent = new SphericalMercator().bbox(x, y, z, false, process.env.EPSG)
```
KlaasH: This is still trying to load the projection from the environment.  It should go below, inside the `map.fromString` callback, and use the projection loaded from the config, i.e. modify the map with
```
result.extent = new SphericalMercator().bbox(x, y, z, false, result.srs)
```
before calling `resolve`.  (I might also vote for a more descriptive variable name than `result`, e.g. `configuredMap`.)

KlaasH: Upon further inspection, that last argument isn't doing anything anyway.  The argument is called `srs` [in the docs](https://github.com/mapbox/sphericalmercator#bboxx-y-zoom-tms_style-srs), but it doesn't actually support arbitrary projections.  It just [checks for the value `'900913'`](https://github.com/mapbox/sphericalmercator/blob/master/sphericalmercator.js#L116) (the [unofficial/non-EPSG id of the web mercator projection](https://en.wikipedia.org/wiki/Web_Mercator#Identifiers)) and converts to that if it matches.  Otherwise it returns WGS84 lat/lon (aka EPSG 4326).

So yeah, the [CartoCSS config uses PROJ4 format](https://cartocss.readthedocs.io/en/latest/mml.html#srs) so if we do need to get `SphericalMercator` to convert the bbox to Web Mercator we'd need to parse the configured SRS and pass the 900913 ID instead.  If we need to support other projections, I think we'd need to add logic to detect the incoming SRS and reproject.

I think we're OK, though--that we can just remove the last argument from this function call and only support lat/lon geometries for now.  Which means we could move it inside the callback, but for now it will work the same in either place.

KlaasH (changes requested):
This is working well and looks good.  A couple more small notes, plus an extended essay on bounding box projections that comes down to "it's pretty much fine".

node_lambnik/src/tiler/package.json
mattdelsordo: Ah, that explains some previous behavior I ended up working around. I think I just assumed this was working because the check is for whether or not NODE_ENV==='production', which is set in a different script.

node_lambnik/src/tiler/src/tiler.js
mattdelsordo: Interesting that the `srs` parameter doesn't actually handle projections in the way I assumed it did. I'll move that line into the callback for future reference.

KlaasH (approved):
:+1:

PR #42, opened 2018-07-05
Feature/extend aws options

### LINKED ISSUE #34: Configure AWS preferences in .env
Allow .env configuration for:
 * Memory
 * ~Timeout~
 * VPCs
 * Security groups

mattdelsordo:
## Overview

This PR extends the configurability of AWS Lambda on deployment by adding more options to `.env`. New options include:
 * `LAMBDA_REGION`: AWS region to which you are deploying
 * `LAMBDA_ROLE`: IAM role given to your lambdas
 * `LAMBDA_TIMEOUT`: Time in seconds before your lambdas time out
 * `LAMBDA_MEMORY`: Memory in MB allocated to your lambdas
 * `LAMBDA_SUBNETS`: VPC Subnets used to configure VPC
 * `LAMBDA_SECURITY_GROUPS`: VPC Security Groups used to configure VPC

The first two options are required on all new deployments, the rest are optional. 

### Notes
Only `LAMBDA_TIMEOUT` and `LAMBDA_MEMORY` are changed on "update" deployments, all other settings can only be changed by deploying a new lambda (or by logging in to the Lambda Management Console).

## Testing Instructions

 * Create a new `.env` file based on the contents of `env-template`.
 * Set up and deploy a new lambda, or update an old one.
 * Change the value of one of the settings variables and re-deploy.
 * Log in to the Lambda Management Console and you should see that the value of that setting has updated.
 

Resolves #34


node_lambnik/env-template
```diff
@@ -10,4 +10,16 @@ PROD_TILEGARDEN_PASSWORD=
...
+
+# OPTIONAL
+# Amount of time in seconds your lambdas will wait before timing out
+#LAMBDA_TIMEOUT=5
+# Memory in MB allocated to your lambda functions
+#LAMBDA_MEMORY=3008
+# VPC Subnets that your lambdas should use (comma separated list)
+#LAMBDA_SUBNETS=subnet1,subnet2,subnet...N
+# VPC Security Groups that your lambdas should use (comma separated list)
+#LAMBDA_SECURITY_GROUPS=group1,group2,group...N
```
KlaasH: I don't think it's totally clear what the `SUBNETS` and `SECURITY_GROUPS` variables are for and how to find values for them.  It might be more detail than we want in the README, so maybe a bit more info in the comments in this file would be the way to go.  E.g. "If you're using RDS, the subnets and security groups should match those of your RDS instance."

KlaasH (commented):
> Only LAMBDA_TIMEOUT and LAMBDA_MEMORY are changed on "update" deployments, all other settings can only be changed by deploying a new lambda.

There's probably no way to make `publish` throw an error if someone tries to publish when they should delete and recreate because they're trying to change something that doesn't get changed on existing functions, but a note in the README about the nuances of publishing and updating would probably be good.

Also, this is making me wonder if we need a script to destroy the published resources (in which case we should add that as another issue).

mattdelsordo:
The closest thing to throwing an error would be the one that pops up if you try to deploy a new lambda with the same `claudia.json` file, but yeah, you wouldn't really run into that if you're just running the publish script over and over. Claudia doesn't seem to complain if you pass flags to it that it doesn't recognize, only if the values for flags it does recognize are malformed. I'll add an issue for making a destroy script, I think it would be useful. That said it would probably a lot easier for a user to log in to AWS and make changes there instead of taking the whole project down just to change a couple of parameters.

KlaasH (approved):
I deleted my stuff and rebuilt and got "Internal server error".  But then I did it again and it worked fine.  I wish I knew what that was about, but hopefully it won't happen again and we won't have to deal with it.

This looks good. Thanks for the additional explanatory comment. :+1:

PR #45, opened 2018-07-06
Feature/filtering support

### LINKED ISSUE #27: Figure out how to make filtering configurable
This is an issue to make a plan. It should result in further issue to implement the plan.

Items to consider:
- See what Windshaft does. They’ve put lots of time into this issue.
- Look at the node-mapnik interface for filtering/datasource config
- Think about what will need to be passed via URL/request and how Claudia.js will handle that

mattdelsordo:
## Overview

This PR adds layer filtering functionality. To quote what I put in the README:

> Map data can be filtered by specifying a query in the form `( QUERY ) as VARIABLE` as the value of `table` for a layer in `map-config.mml`. Different combinations of layers can be fetched by adding `?layers=layer1,layer2,layer...N` as a query string to a tile or UTF grid URL. Make sure that all necessary columns are present in the query. For example, if your geometry column is called `geom`, and you want to display a UTF grid of content from the column `name` filtered by the content of the column `type`, your query would look something like `SELECT geom,name,type FROM table WHERE type='something'`. If no layers are specified in the query string, all layers are displayed in the same tile.

### Demo

image
image

### Notes

There's a persisting shearing issue in tiles with large numbers of markers, this will be addressed in a subsequent PR.

## Testing Instructions

 * Run the development server as per usual.
 * Open up the demo webpage in `/src/demo`. It already has several example filters enabled on it.
 * To add a filer to the demo, you have to alter the following:
   * `src/tiler/src/config/map-config.mml`: Add a new layer with a unique name and a new query as the value of `table`. For your convenience, here's the values of `owner` that I left out of the demo: 
     * PRIVPWDMAINT
     * CHEL
     * PRIVMAND
     * STREET
     * BENS
     * HVF
     * PIDC
     * LML
     * SPR
     * AUTH
     * ABT
    * `src/tiler/src/config/style.mss`: Add a new style class with the same name as the layer you made and style it however you want.
    * `src/demo/pwd_inlets/index.html`: In the `#layer-selector` span, add a new input/label cluster for your layer. Then, in the script section, add a new `createLayer('$NAME_OF_YOUR_LAYER')`.
  * Reload the demo page and check your layer's box to see if it is properly served up.

Resolves #27

### DIFF
--- a/node_lambnik/README.md
+++ b/node_lambnik/README.md
@@ -15,6 +15,9 @@
 ### Configuration & Styling
 [Tilegarden uses CartoCSS to specify map styles.](https://carto.com/docs/carto-engine/cartocss/) Edit `src/tiler/src/config/map-config.mml` to specify your map configuration, including a reference any CartoCSS `.mss` files. [See here](https://cartocss.readthedocs.io/en/latest/mml.html) for Carto's `.mml` specification.
 
+### Filtering
+Map data can be filtered by specifying a query in the form `( QUERY ) as VARIABLE` as the value of `table` for a layer in `map-config.mml`. Different combinations of layers can be fetched by adding `?layers=layer1,layer2,layer...N` as a query string to a tile or UTF grid URL. All specified layers are shown by default.
+
 ### Debugging
  The local development server exposes a websocket to a node inspector that can be attached to your IDE of choice to step through and debug your code. Here are instructions on how to do so using Google Chrome's Dev Tools:
   * Start the development server with `./scripts/server`.
--- a/node_lambnik/README.md
+++ b/node_lambnik/README.md
@@ -6,17 +6,17 @@
  * Run `./scripts/server` to run the development server, which will be accessible on port **9001**, not 3000 as listed by the output from claudia-local-api. Several example endpoints have been defined in `src/api.js`.
  	* Display usage info: `/`
  	* Generate a map tile: `/tile/{z}/{x}/{y}.png` serves tiles.
- 	* Generate a UTF grid: `/grid/{z}/{x}/{y}`
- * To publish, make sure you have specified valid AWS credentials and have listed the proper database credentials in a `.env` file in the root of the project. The format for this file can be copied from `env-template`. You can then run `./scripts/publish --new` to publish to AWS Lambda.
-   * For subsequent deployments of the same lambdas, leave off the `--new` tag.
-   * **NOTE:** Only `LAMBDA_TIMEOUT` and `LAMBDA_MEMORY` are changed on "update" deployments, all other settings must be changed by logging in to the Lambda Management Console or by deploying a new project.
+ 	* Generate a UTF grid: `/grid/{z}/{x}/{y}?utfFields=field1,field2,field...N` etc, where a `field` is a column in your data table.
+ * To publish, make sure you have specified valid AWS credentials and have listed the proper database credentials in a `.env` file in the root of the project. The format for this file can be copied from `env-template`. You can then run `./scripts/publish` to publish to AWS Lambda.
+    * For subsequent deployments of the same lambdas, leave off the `--new` tag.
+    * **NOTE:** Only `LAMBDA_TIMEOUT` and `LAMBDA_MEMORY` are changed on "update" deployments, all other settings must be changed by logging in to the Lambda Management Console or by deploying a new project.
  * API Gateway has trouble serving images, so you need to configure some sort of proxy (e.g. using a CloudFront distribution) that sets an `Accept:image/png` header on all lambda requests.
  
 ### Configuration & Styling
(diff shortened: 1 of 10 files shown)


node_lambnik/README.md
```diff
@@ -6,15 +6,18 @@
...
  
 ### Configuration & Styling
 [Tilegarden uses CartoCSS to specify map styles.](https://carto.com/docs/carto-engine/cartocss/) Edit `src/tiler/src/config/map-config.mml` to specify your map configuration, including a reference any CartoCSS `.mss` files. [See here](https://cartocss.readthedocs.io/en/latest/mml.html) for Carto's `.mml` specification.
 
+### Filtering
+Map data can be filtered by specifying a query in the form `( QUERY ) as VARIABLE` as the value of `table` for a layer in `map-config.mml`. Different combinations of layers can be fetched by adding `?layers=layer1,layer2,layer...N` as a query string to a tile or UTF grid URL. Make sure that all necessary columns are present in the query, for example, if your geometry column is called `geom`, and you want to display a UTF grid of content from the column `name` filtered by the content of the column `type`, your query would look something like `SELECT geom,name,type FROM table WHERE type='something''`. If no layers are specified in the query string, all layers are displayed in the same tile.
```
KlaasH: Copy notes: I would start a new sentence at "For example", and there's an extra quote on the example query.

node_lambnik/src/demo/pwd_inlets_index.html
```diff
@@ -35,39 +62,71 @@ <h3>PWD Inlets</h3>
...
             }
+        }
+
+        createLayer('PWD');
+        createLayer('STATE');
+        createLayer('PARK');
+        createLayer('AIRPRT');
+        createLayer('LM');
+        createLayer('FEDERAL');
+        createLayer('PRIV');
```
KlaasH: It's just code on a demo page, but the repetition here and the indirectness above are bothering me a little.  The latter especially, since I would expect `createLayer` to return the layer rather than modify an object from outside its scope.

I would vote for changing the function to return the layer and changing the object creation to:
```javascript
var layerKeys = ['PWD', 'STATE', 'PARK', 'AIRPRT', 'LM', 'FEDERAL', 'PRIV']
const layers = Object.assign(...layerKeys.map(key => ({[key]: createLayer(key)})));
```
(and using `layers` below rathar than `layerKeys`. more accurately descriptive)
That's an annoying number of closing parens with a curly brace thrown in for extra unreadability, but it has the advantage that it makes the object directly rather than as a side-effect.  And it's short.

node_lambnik/src/tiler/src/tiler.js
```diff
@@ -17,21 +17,42 @@ const TILE_WIDTH = 256
...
+const filterLayers = (xmlString, layers) => {
+    // Only filter if there ARE layers to filter on
+    if (layers && layers.length > 0) {
+        return xmlString.replace(
+            /(<Layer name="(.+)")>/g,
+            (fullMatch, beginning, layerName) => {
+                if (!layers.includes(layerName)) {
+                    return `${beginning} status="0">`
+                }
+                return fullMatch
```
KlaasH: I think trying to parse markup with regexes is always a dangerous game, and the fact that this very simple attribute addition is not particularly straightforward to read and already seems kind of brittle (can `Layer` have other attributes besides `name`?  If so, this would stop working.) is a decent illustration of why.  I think we should find a (hopefully pretty lightweight, since our needs are quite modest) package to load this as XML and let us traverse it and set attributes as such, then write it back out to a string.

node_lambnik/src/tiler/src/tiler.js
```diff
@@ -17,21 +17,42 @@ const TILE_WIDTH = 256
...
-const createMap = (z, x, y) => {
+const createMap = (z, x, y, layers) => {
     // Create a webmercator map with specified bounds
     const map = new mapnik.Map(TILE_WIDTH, TILE_HEIGHT)
     map.bufferSize = 64
     map.extent = new SphericalMercator().bbox(x, y, z, false, process.env.EPSG)
 
     // Load map specification from xml string
     return readFile(path.join(__dirname, 'map-config.xml'), 'utf-8')
+        .then(xml => filterLayers(xml, layers))
```
KlaasH: Is it really true that layers can't be dropped after the map object is made?  It certainly [looks that way](http://mapnik.org/documentation/node-mapnik/3.6/#Map), but it seems like such obviously useful functionality...

KlaasH (changes requested):

node_lambnik/src/tiler/src/tiler.js
mattdelsordo: Yeah, I agree, it seems like a weird design choice to me. [I found this issue in node-mapnik](https://github.com/mapnik/node-mapnik/issues/742) where someone expresses a need for the same functionality, and Dane Springmeyer (who's the chief contributor to the project) said the following:
> This is a known limitation that I don't think should be changed. I don't think an API that allows mutating maps in place is within scope. I recommend thinking about map instances as best created once and then rendered one or more times and never modified again.

node_lambnik/src/tiler/src/tiler.js
mattdelsordo: I agree, I've realized when writing tests that it already doesn't work even in the simple use cases we've been demoing it on. I'll go swap this out for an XML library, we might need one for YAML too.

KlaasH (approved):
Looks great! :+1:

PR #50, opened 2018-07-09
Feature/cleanup scripts

### LINKED ISSUE #44: Write destroy script
One script should be able to remove all lambda/API Gateway/CloudFront functionality for a clean replacement if core configuration options need to change.

mattdelsordo:
## Overview

This PR adds two scripts, `destroy`, which destroys the currently deployed AWS resources, and `clean`, which deletes local development environment-created files (such as `node_modules/` and downed Docker containers).

### Demo
```
WARNING: you are about to destroy all your deployed resources!
If this is what you want, type 'yes' to continue: yes

Starting nodelambnik_database_1 ... done
yarn run v1.5.1
$ claudia destroy
{ AccessDenied: User: arn:aws:iam::279682201306:user/mdelsordo is not authorized to perform: iam:DeleteRolePolicy on resource: role lambnik-tiler-dev
    at Request.extractError (/home/tiler/node_modules/aws-sdk/lib/protocol/query.js:47:29)
    at Request.callListeners (/home/tiler/node_modules/aws-sdk/lib/sequential_executor.js:105:20)
    at Request.emit (/home/tiler/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
    at Request.emit (/home/tiler/node_modules/aws-sdk/lib/request.js:683:14)
    at Request.transition (/home/tiler/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/home/tiler/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /home/tiler/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/home/tiler/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/home/tiler/node_modules/aws-sdk/lib/request.js:685:12)
    at Request.callListeners (/home/tiler/node_modules/aws-sdk/lib/sequential_executor.js:115:18)
  message: 'User: arn:aws:iam::279682201306:user/mdelsordo is not authorized to perform: iam:DeleteRolePolicy on resource: role lambnik-tiler-dev',
  code: 'AccessDenied',
  time: 2018-07-09T15:38:02.402Z,
  requestId: '10b545b1-838e-11e8-b203-1933c93aab81',
  statusCode: 403,
  retryable: false,
  retryDelay: 26.99817764686927 }
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c claudia destroy
Directory: /home/tiler
Output:
".
info If you think this is a bug, please open a bug report with the information provided in "/home/tiler/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Starting nodelambnik_database_1 ... done
./scripts/destroy: line 25: src/tiler/.api-id: No such file or directory
aws_cloudfront_distribution.tilegarden_test: Refreshing state... (ID: E1SLZVWI7WGQ9N)
aws_cloudfront_distribution.tilegarden_test: Destroying... (ID: E1SLZVWI7WGQ9N)
aws_cloudfront_distribution.tilegarden_test: Still destroying... (ID: E1SLZVWI7WGQ9N, 10s elapsed)
...
...
aws_cloudfront_distribution.tilegarden_test: Still destroying... (ID: E1SLZVWI7WGQ9N, 18m20s elapsed)
aws_cloudfront_distribution.tilegarden_test: Still destroying... (ID: E1SLZVWI7WGQ9N, 18m30s elapsed)
aws_cloudfront_distribution.tilegarden_test: Destruction complete after 18m34s

Destroy complete! Resources: 1 destroyed.
```

### Notes

* The final step of Claudia's `destroy` command is to delete all IAM roles associated with the (at this point deleted) lambda. This is not super great for our use case as the `lambnik-tiler-dev` role is shared between various AWS resources. If you don't have the permissions to delete roles (which I don't) the command will fail, but will still have removed the Lambda and API Gateway instances, so the end result is the desired functionality. The trouble comes in when you DO have the permissions to delete roles, and I'm not quite sure how to work around it as `claudia destroy` lacks a ton of configuration options.
* Destroying a CloudFront distribution takes a REALLY long time. I decided to leave the command line output in (rather than running the process in detached mode) so that you can at least get progress updates and not forget about the process before it's complete.

#### (assuming you have already deployed a Tilegarden project and have `claudia.json` and `.api-id` files locally)
 * `destroy`
    * Run `./scripts/destroy`. It will prompt you for confirmation before beginning the process.
    * Wait for the various resources to be removed.
    * All AWS services should be gone along with local config files such as `tiler/claudia.json` and `terraform/terraform.tfstate`.
 * `clean`
    * Run `./scripts/clean`.
    * `tiler/bin/`, `tiler/node_modules/`, and all docker containers should now be removed.

Resolves #44

### DIFF
--- a/node_lambnik/env-template
+++ b/node_lambnik/env-template
@@ -16,8 +16,8 @@ PROD_TILEGARDEN_USER=
 LAMBDA_REGION=us-east-1
 
 ## OPTIONAL ##
-# ARN of role associated with this lambda function
-#LAMBDA_ROLE=arn:aws:iam:1237192731893/role:etc
+# name of role associated with this lambda function
+#LAMBDA_ROLE=role-name
 # Amount of time in seconds your lambdas will wait before timing out
 #LAMBDA_TIMEOUT=5
 # Memory in MB allocated to your lambda functions
--- a/node_lambnik/env-template
+++ b/node_lambnik/env-template
@@ -14,9 +14,10 @@ PROD_TILEGARDEN_USER=
 # Function config information
 ## REQUIRED ##
 LAMBDA_REGION=us-east-1
-LAMBDA_ROLE=role
 
 ## OPTIONAL ##
+# ARN of role associated with this lambda function
+#LAMBDA_ROLE=arn:aws:iam:1237192731893/role:etc
 # Amount of time in seconds your lambdas will wait before timing out
 #LAMBDA_TIMEOUT=5
 # Memory in MB allocated to your lambda functions
--- a/node_lambnik/scripts/clean
+++ b/node_lambnik/scripts/clean
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+#!/bin/bash
+
+# Most reliable way to get the path for this script.
+# h/t: https://stackoverflow.com/questions/192292/bash-how-best-to-include-other-scripts/12694189#12694189
+DIR="${BASH_SOURCE%/*}"
+if [[ ! -d "$DIR" ]];
+then
+    DIR="$PWD"
+fi
+
+function usage() {
+    echo -n "Usage: $(basename "$0")
+Clean up unused Docker resources to free disk space.
+"
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+    	# Remove build artifacts from local machine
+    	docker-compose run tiler rm -rf bin/ node_modules/
+
+        # Stop containers and shut down the network
+        docker-compose down
+
+        # Remove unused containers, images, and volumes
+        docker images -qf dangling=true | xargs -r docker rmi
+        docker volume ls -qf dangling=true | xargs -r docker volume rm
+        docker ps -a | grep Exit | cut -d ' ' -f 1 | xargs -r docker rm
+
+        docker-compose rm
+    fi
+fi
--- a/node_lambnik/scripts/clean
+++ b/node_lambnik/scripts/clean
@@ -1,14 +1,4 @@
-#!/usr/bin/env bash
-
 #!/bin/bash
-
-# Most reliable way to get the path for this script.
-# h/t: https://stackoverflow.com/questions/192292/bash-how-best-to-include-other-scripts/12694189#12694189
-DIR="${BASH_SOURCE%/*}"
-if [[ ! -d "$DIR" ]];
-then
-    DIR="$PWD"
-fi
 
 function usage() {
     echo -n "Usage: $(basename "$0")
@@ -20,8 +10,8 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
-    	# Remove build artifacts from local machine
-    	docker-compose run tiler rm -rf bin/ node_modules/
+        # Remove build artifacts from local machine
+        docker-compose run tiler rm -rf bin/ node_modules/
 
         # Stop containers and shut down the network
         docker-compose down --volumes
--- a/node_lambnik/scripts/clean
+++ b/node_lambnik/scripts/clean
@@ -12,7 +12,7 @@ fi
 
 function usage() {
     echo -n "Usage: $(basename "$0")
-Clean up unused Docker resources to free disk space.
+Destroy build artifacts and Docker containers for a fresh environment
 "
 }
 
@@ -24,13 +24,10 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     	docker-compose run tiler rm -rf bin/ node_modules/
 
         # Stop containers and shut down the network
-        docker-compose down
+        docker-compose down --volumes
 
-        # Remove unused containers, images, and volumes
+        # Remove any dangling volumes
         docker images -qf dangling=true | xargs -r docker rmi
         docker volume ls -qf dangling=true | xargs -r docker volume rm
-        docker ps -a | grep Exit | cut -d ' ' -f 1 | xargs -r docker rm
-
-        docker-compose rm
     fi
 fi
--- a/node_lambnik/scripts/destroy
+++ b/node_lambnik/scripts/destroy
@@ -12,6 +12,7 @@ Options:
 
 function main() {
     echo "WARNING: you are about to destroy all your deployed resources!"
+    printf "That includes:\n - Associated AWS Lambda functions\n - Associated API Gateways\n - Associated IAM roles (automatically generated or no)\n - Associated CloudFront distributions\n"
     read -p "If this is what you want, type 'yes' to continue: " -r
     echo
     if [[ $REPLY =~ ^[yY][eE][sS]$ ]]
--- a/node_lambnik/scripts/destroy
+++ b/node_lambnik/scripts/destroy
@@ -16,8 +16,15 @@ function main() {
 	echo
 	if [[ $REPLY =~ ^[yY][eE][sS]$ ]]
 	then
-		docker-compose run tiler yarn destroy;
-		docker-compose run terraform destroy -auto-approve;
+		# This command ALSO tried to destroy your lambda's IAM role.
+		# It will probably fail to do so, but it will delete the lambda
+		# and API Gateway FIRST, so it's fine.
+		docker-compose run tiler yarn destroy || true;
+		docker-compose run tiler rm claudia.json;
+
+		docker-compose run -e TF_VAR_source_id=$(<src/tiler/.api-id) terraform destroy -auto-approve;
+		docker-compose run --entrypoint rm terraform -rf .terraform terraform.tfstate terraform.tfstate.backup;
+
 	else
 		echo "Aborting destroy"
 	fi
--- a/node_lambnik/scripts/destroy
+++ b/node_lambnik/scripts/destroy
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+set -e
+
+function usage() {
+    echo -n "Usage: $(basename "${0}")
+Removes all published resources from AWS
+Options:
+    --help      Display this help text
+"
+}
+
+function main() {
+	echo "WARNING: you are about to destroy all your deployed resources!"
+	read -p "If this is what you want, type 'yes' to continue: " -r
+	echo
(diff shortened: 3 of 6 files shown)


node_lambnik/scripts/clean
```diff
@@ -0,0 +1,36 @@
...
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+    	# Remove build artifacts from local machine
+    	docker-compose run tiler rm -rf bin/ node_modules/
+
+        # Stop containers and shut down the network
+        docker-compose down
```
KlaasH: The line to delete dangling volumes below will catch them, but I still think it makes sense to include the `--volumes` option here.

node_lambnik/scripts/clean
```diff
@@ -0,0 +1,36 @@
...
+
+        # Remove unused containers, images, and volumes
+        docker images -qf dangling=true | xargs -r docker rmi
+        docker volume ls -qf dangling=true | xargs -r docker volume rm
+        docker ps -a | grep Exit | cut -d ' ' -f 1 | xargs -r docker rm
```
KlaasH: Removing dangling images and volumes is good, but this line too aggressive.  This is run from the host, so if someone has other docker projects going, they might not be pleased to find that this has reached out and deleted all their containers.

There's a `docker-compose ps` command that shows only containers connected to the docker-compose config, though it doesn't seem to show stopped terraform containers.  There's also `docker-compose rm -fv`, which does delete stopped single-run containers.  But `docker-compose down` does, too.  I'm not sure this needs to be here at all.  Is there a category of containers that's getting missed by `down`?

node_lambnik/scripts/clean
```diff
@@ -0,0 +1,36 @@
...
+# h/t: https://stackoverflow.com/questions/192292/bash-how-best-to-include-other-scripts/12694189#12694189
+DIR="${BASH_SOURCE%/*}"
+if [[ ! -d "$DIR" ]];
+then
+    DIR="$PWD"
+fi
+
+function usage() {
+    echo -n "Usage: $(basename "$0")
+Clean up unused Docker resources to free disk space.
```
KlaasH: This doesn't just clean up unused resources, it destroys everything.  One possibility would be to use options to define different levels (just removing old single-use containers and dangling volumes; that plus removing the bin and node_modules directories and any other artifacts along those lines; full destruction of local resources).  But I think for now, we could just do full destruction, since it's not hard to rebuild from scratch.  So this usage message should adjust to reflect that.

node_lambnik/scripts/destroy
```diff
@@ -0,0 +1,45 @@
...
+		# Destroy deployed resources
+		# This command ALSO tried to destroy your lambda's IAM role.
+		# It will probably fail to do so, but it will delete the lambda
+		# and API Gateway FIRST, so it's fine.
+		docker-compose run tiler yarn destroy || true;
+		docker-compose run -e TF_VAR_source_id=$(<src/tiler/.api-id) terraform destroy -auto-approve;
+
+		# Remove local config files
+		docker-compose run tiler rm claudia.json .api-id;
+		docker-compose run --entrypoint rm terraform -rf .terraform terraform.tfstate terraform.tfstate.backup;
```
KlaasH: These two commands can just be done on the host, since they're clearing files from linked directories.  No need to spin up a container just to run `rm`.

KlaasH (changes requested):
Looks good over all.  A few notes below.

Sometimes I wonder why `--rm` isn't the default behavior for `docker-compose run`, with `--keep` being an override.  There's probably a reason, but it certainly does create cruft...

node_lambnik/scripts/destroy
mattdelsordo: The issue is that the files get created by the containers so the host user doesn't have the permissions to delete them. I felt it was easier to run the commands from the containers than to have the user run the script with `sudo` every time.

node_lambnik/scripts/clean
mattdelsordo: Ah, alright, my intent with this was to try and remove outdated images, which `docker-compose down` doesn't seem to do, but this seems like overkill for that purpose. I'll rework this into something less extreme.

node_lambnik/scripts/destroy
KlaasH: Oh right!  I set up user namespacing, which is a global setting for docker on the host machine, so I don't have those problems anymore.  It's weird and complicated enough that it probably doesn't make sense to add it to the README as expected developer setup, which means I should probably turn it back off so I'm not blind to that issue...

node_lambnik/scripts/clean
KlaasH: Outdated in the sense that they're not referenced by the current `docker-compose.yml`, or some other way?  I ask mainly out of curiosity, to understand the gaps.  Containers that changed name or were removed from the config file are the one category I can think of.

The names all get prefixed with `nodelambnik_`, so we could filter `docker ps -a` by that...

mattdelsordo: Dangling images, I mean. I'm pretty sure that's what I'm looking for but I'm still sort of wrapping my head around all the resources Docker creates.

node_lambnik/scripts/clean
```diff
@@ -0,0 +1,33 @@
...
+Destroy build artifacts and Docker containers for a fresh environment
+"
+}
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+    if [ "${1:-}" = "--help" ]; then
+        usage
+    else
+    	# Remove build artifacts from local machine
+    	docker-compose run tiler rm -rf bin/ node_modules/
```
KlaasH: The indentation of these two lines looks the same as the ones below in my editor but different on GitHub.  Turns out that's because there are a couple tabs here, which my editor is rendering with width 4 while GitHub gives them width 2.  Azavea style calls for spaces.

node_lambnik/scripts/destroy
```diff
@@ -0,0 +1,45 @@
...
+function usage() {
+    echo -n "Usage: $(basename "${0}")
+Removes all published resources from AWS
+Options:
+    --help      Display this help text
+"
+}
+
+function main() {
+	echo "WARNING: you are about to destroy all your deployed resources!"
```
KlaasH: This file somehow ended up with mostly tab spacing as well.  And with a bunch of unnecessary line-ending semicolons.

node_lambnik/scripts/destroy
```diff
@@ -0,0 +1,45 @@
...
+	read -p "If this is what you want, type 'yes' to continue: " -r
+	echo
+	if [[ $REPLY =~ ^[yY][eE][sS]$ ]]
+	then
+
+		# Destroy deployed resources
+		# This command ALSO tried to destroy your lambda's IAM role.
+		# It will probably fail to do so, but it will delete the lambda
```
KlaasH: This is a problem.
First, because it shows that we're still relying on hand-rolled AWS entities that aren't provisioned by the provisioning script, nor even documented, currently.
Second, because it means that if someone with more permissions runs it, they could delete the resources that we're relying on but don't have automatic provisioning for.
...which is what just happened.  Turns out I was running with my `research` AWS profile which has `AdministratorAccess`, so the role we've been using, `lambnik-tiler-dev`, is gone now.  We should probably restore that one by hand then figure out how to add role creation to the publish script before we merge this.  (Can Claudia create the necessary role?  It would seem weird for it to be deleting it by default if it's not able to create it.)

KlaasH (changes requested):
This might not actually need changes (except the whitespace and semicolons), but before we merge a thing that tries to delete the IAM role, we need a new issue to either get claudia.js to provision it the or add it to the Terraform provisioning setup.

node_lambnik/scripts/destroy
mattdelsordo: Oh crap, I should have asked you about this issue right when I came across it, this is really my bad. What's happening now is that you can optionally configure a role in the `.env`, which I've been doing this whole time since one already existed to do the things lambnik needed to, and because I didn't have the permissions to create new roles. If you leave that off, Claudia will automatically create a new role for your lambda (assuming you have the permissions to do so).

I was able to find a gitter conversation on the Claudia repo in which some people discussed adding a tag to exclude the IAM role from being touched by `claudia destroy` for exactly this use case. Unfortunately it doesn't seem like anyone ever actually implemented it. What do you think the best course of action would be to make sure this script is safe to use?

node_lambnik/scripts/clean
mattdelsordo: Huh, that's weird, my editor is configured for spaces, I wonder how those got in there.

node_lambnik/env-template
```diff
@@ -14,9 +14,10 @@ PROD_TILEGARDEN_USER=
 # Function config information
 ## REQUIRED ##
 LAMBDA_REGION=us-east-1
-LAMBDA_ROLE=role
 
 ## OPTIONAL ##
+# ARN of role associated with this lambda function
+#LAMBDA_ROLE=arn:aws:iam:1237192731893/role:etc
```
KlaasH: Does it need to be the full ARN?  Wasn't it working with just the name?

node_lambnik/env-template
```diff
@@ -14,9 +14,10 @@ PROD_TILEGARDEN_USER=
 # Function config information
 ## REQUIRED ##
 LAMBDA_REGION=us-east-1
-LAMBDA_ROLE=role
 
 ## OPTIONAL ##
+# ARN of role associated with this lambda function
```
KlaasH: It's worth adding a note here about the overzealous-deletion bug.  Which means we should include an issue number on the claudia.js repo.  Which means we should report the issue.

KlaasH (approved):
Working! Yay!
I opened #53 as a record of what got us past the role permissions issues (and because we need a permissions section in the README regardless).  One suggestion and one question below, :+1: pending those.

node_lambnik/env-template
mattdelsordo: I'm on it, I was just now making sure I could reproduce the not-writing-to-file bug.

mattdelsordo: I've also added a line re: the overzealous deletion to the warning message in `destroy`.

PR #56, opened 2018-07-13
Feature/line+polygon support

### LINKED ISSUE #54: Enable addition of variable amounts of tables to the development database
You currently have to manually edit a file to load one .sql file worth of data into the development database.

### LINKED ISSUE #55: Support all geographic features
Lines and polygons should be drawn as readily as points currently are.

mattdelsordo:
## Overview

This PR demonstrates Tilegarden's ability to serve tiles containing lines and polygons in addition to points. To facilitate testing this on multiple data sets, the script that sets up the development database has been modified to import all gzipped `.sql` files and all zipped shapefiles in `/data` automatically.

### Demo

image
image
image

### Notes

The tile shearing issue is even more noticeable when rendering polygons, that is going to be next up to be fixed.

## Testing Instructions

 * Run the server normally with `./scripts/update` and `./scripts/server`.
 * Open `/src/demo/datatype-demo.html`.

Resolves #54, resolves #55


node_lambnik/src/database/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM quay.io/azavea/postgis:2.3-postgres9.6-slim
+
+RUN apt-get update
+RUN apt-get install postgis -y
```
mmcfarland, line 4: This line is a little curious since postgis should be available as built in the base image: https://github.com/azavea/docker-postgis/blob/master/Dockerfile-slim.template

Is it that the shp2pgsql tool isn't available via base?

node_lambnik/scripts/update
```diff
@@ -21,9 +21,10 @@ then
 
         # Launch database container so we can populate it.
         docker-compose up -d database
+        docker-compose exec database bash -c "./updatedb"
```
mmcfarland: I don't seem to be able to run this the first time with success and have it load data.  After running `server` and building the containers, update does appear to reload data correctly.

node_lambnik/src/demo/datatype-demo.html
```diff
@@ -0,0 +1,108 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset="utf-8">
+    <title>Datatype Demo</title>
+
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.7/leaflet.css" rel="stylesheet" type="text/css" />
```
mmcfarland: This is likely of little consequence, especially as a demo, but I'd recommend using the much more recent 1.x series of leaflet.

node_lambnik/src/tiler/src/config/style.mss
```diff
@@ -1,38 +1,23 @@
...
 
-#FEDERAL {
-    marker-fill: red;
-    marker-width: 4;
+#street_centerline{
+    marker-fill: black;
+    marker-width: 2;
     marker-allow-overlap: true;
+    line-color: red;
+    polygon-fill: orange;
```
mmcfarland: Probably want to drop this fill style for a line dataset, which shouldn't have an area to fill in principle.

node_lambnik/scripts/update
```diff
@@ -21,9 +21,6 @@ then
 
         # Launch database container so we can populate it.
         docker-compose up -d database
-
-        # Load data fixtures.
-        ./scripts/updatedb
-
+        docker-compose exec database bash -c "./updatedb"
```
mmcfarland, line 24: There seems to be some race condition here that isn't accounted for in the updatedb script which checks the databases availability.  If I've previously built the containers via `docker-compose up` or `./scripts/server`, this script is successful.  However, after a `down` & `clean`, running `update` won't load the data.

mmcfarland (commented):
Got everything working locally, but didn't publish as I never set up appropriate .env values.

node_lambnik/src/database/Dockerfile
mattdelsordo: Exactly, I couldn't locate the package except as part of postgis.

node_lambnik/src/demo/datatype-demo.html
mattdelsordo: I left it at 0.7.x because I couldn't get 1.0 to work with utf grids and didn't want to waste time on it at the time, but I probably should update that.

node_lambnik/src/database/Dockerfile
mattdelsordo: [According to this](https://gis.stackexchange.com/questions/148524/i-have-not-found-shp2pgsql-in-postgis-installation) shp2pgsql only comes with postgis-client, which isn't on the image. I'll use that instead of the entirety of postgis.

mattdelsordo: Scratch that, [it looks like the supported way of installing shp2pgsql is to do exactly what I did](https://trac.osgeo.org/postgis/wiki/UsersWikiPostGIS23UbuntuPGSQL96Apt#Install).

node_lambnik/scripts/update
mattdelsordo: I for the life of me cannot seem to replicate this problem (not to say I doubt it's an issue). @KlaasH , does this happen on your machine as well?

node_lambnik/src/database/Dockerfile
mmcfarland: Ah, great then.  I was afraid we were installing potentially mixed versions of the database extension.

node_lambnik/scripts/update
mmcfarland: I think this is tangential enough, and with a workaround, to move forward.  Can you create an issue to follow up separately?

mattdelsordo: Sure thing.

mmcfarland (approved):

PR #59, opened 2018-07-16
Upgrade Claudia to 5.0.1

mattdelsordo:
## Overview

This PR upgrades Claudia to v5.0.1, which includes a fix for the bug that would cause you to delete shared IAM roles even if they were pre-specified: https://github.com/claudiajs/claudia/issues/162#issuecomment-404856082

## Testing Instructions

 * Deploy a project to AWS that shares it's IAM role with another resource.
 * Run `./scripts/destroy`. After it completes, the role should still exist on AWS.

### DIFF
--- a/node_lambnik/scripts/destroy
+++ b/node_lambnik/scripts/destroy
@@ -12,7 +12,7 @@ Options:
 
 function main() {
     echo "WARNING: you are about to destroy all your deployed resources!"
-    printf "That includes:\n - Associated AWS Lambda functions\n - Associated API Gateways\n - Associated IAM roles (automatically generated or no)\n - Associated CloudFront distributions\n"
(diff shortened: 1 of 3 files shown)


KlaasH (approved):
:+1: 

Nice turnaround from the author.  Though I see he kept the "save then format" setup with both building config objects that are almost but not quite the same.  Not sure why, but this does the trick for us.

One note: the warning in [`scripts/destroy`](https://github.com/azavea/tilegarden/blob/671a1ebfa385e6ccfc09915b52fbba09b3138808/node_lambnik/scripts/destroy#L15) should be changed from "Associated IAM roles (automatically generated or no)" to "Associated IAM roles (if automatically generated)"

PR #60, opened 2018-07-17
Feature/fix tile shearing

### LINKED ISSUE #43: Identify/fix tile shearing issue
Matt suggested it may be a mapnik buffer size issue.

mattdelsordo:
## Overview

This PR fixes the warping that was taking place when rendering tiles, which ended up being the result of data being mis-projected:
 * The target srs was not specified properly in `map-config.mml`. Tiles for the demo had to end up in web mercator but they were being projected as WGS84, which was causing the warping. If the source and target srs strings both match the data in the database and the desired output projection respectively, mapnik will automatically re-project the data.
 * Bounding boxes for individual tiles were being calculated in WGS84, which meant that even when tiles were being properly projected as web mercator they were being cropped at the wrong latitude/longitude. I removed the bounding box calculation dependency and wrote a new version that can calculate bounding boxes for any target projection.

### Demo
image

## Testing Instructions

 * Run `./scripts/clean`, `./scripts/update`, and `./scripts/server` to start a new development server with the updated data from this PR. I included the `street_centerlines` data set in it's original projection in order to demonstrate mapnik's reprojection capabilities but it won't get added to the database if a similarly named table already exists.
* Open `src/demo/datatype-demo.html` to see the server serve properly projected content.

Resolves #43


node_lambnik/src/tiler/src/util/bounding-box.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Math I don't understand was borrowed from
+ * https://github.com/mapbox/sphericalmercator/blob/e33a8c8bf852ea7a7c57d1f9001b581fce4b6590/sphericalmercator.js
+ */
+
+import mapnik from 'mapnik'
+
+const R2D = 180 / Math.PI
+
+const pxToWgs84Ll = (px, zoom, tileSize) => {
```
KlaasH: I see why you pulled this out of SphericalMercator, since that only supports the two projections and here we can take advantage of having Mapnik available to support any projection.  I think, since this is sort of a black box, a bit more narrative about what it's meant to be doing and why you extracted this particular bit of impenetrable math would be useful.  That way future people will 1) be able to read that and not have to spend any time pondering what this is aiming at and 2) have enough context to evaluate, if something doesn't seem to be working right, whether the problem might be coming from here.  For example, citing line line numbers in the original would let someone see whether those lines have been changed since this was copied and whether the change is significant for our purposes.

Also, the function name here ends with `Ll`, but that's for "lower left" and this is just a general-purpose pixel-to-coordinates converter, right?  Not specific to one bound?

mattdelsordo: Will do re: part A. I think there's a bit of a mixup because the original code uses "ll" to mean both "lower left" and "LonLat", I'll change the name so it makes more sense.

node_lambnik/src/tiler/tests/bounding-box.test.js
```diff
@@ -0,0 +1,58 @@
...
+            2686442.41811553,
+            220380.94455046,
+            2698403.39521981,
+            233012.37911430
+        ]
+
+        bbox(testZoom, testX, testY, tileSize, projection).forEach((val, i) => expect(val).toBeCloseTo(target[i], 1))
+    })
+
+    test('Properly use TMS scheme', () => {
```
KlaasH: Aside from the name, this is identical to the first test.

KlaasH (approved):
This is great.  Making any data source projection "just work" is really nice.  See one note below.  :+1: with the identical test differentiated or removed.

node_lambnik/src/tiler/tests/bounding-box.test.js
mattdelsordo: Ah, the original bbox calculation algorithm had support for TMS map tiles, which I removed for simplicity since we don't have any API option to enable that (yet). Looks like I forgot to remove the half-written test. Do you think support for TMS tiles would be a useful feature to implement down the line?

PR #69, opened 2018-07-20
Feature/remove lfs

### LINKED ISSUE #68: LFS has already hit bandwidth cap

mattdelsordo:
## Overview

Using git-lfs to store development data sets cause us to hit GitHub's bandwidth cap very quickly. This PR removes lfs from the project and replaces it with a script that downloads data as `.zip`s directly from OpenDataPhilly.

## Testing Instructions

 * Clear out `/node_modules/data/`.
 * Run `./scripts/fetch-data.sh`.
 * `./scripts/update` should now work as usual.

Resolves #68

### DIFF
--- a/node_lambnik/scripts/fetch-data.sh
+++ b/node_lambnik/scripts/fetch-data.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+# Fetches development data sets from OpenDataPhilly
+curl 'https://phila-gisdata.s3.amazonaws.com/ODP/STR_Centerline.zip' --output data/STR_Centerline.zip
+curl 'http://gis.phila.gov/gisdata/ODP/PWD_Parcels.zip' --output data/PWD_Parcels.zip
+curl 'https://phl.carto.com/api/v2/sql?q=SELECT+*+FROM+inlets&filename=inlets&format=shp&skipfields=cartodb_id' --output data/inlets.zip
--- a/node_lambnik/scripts/fetch-data.sh
+++ b/node_lambnik/scripts/fetch-data.sh
@@ -1,6 +1,14 @@
 #!/usr/bin/env bash
 
 # Fetches development data sets from OpenDataPhilly
-curl 'https://phila-gisdata.s3.amazonaws.com/ODP/STR_Centerline.zip' -L -o data/STR_Centerline.zip
-curl 'http://gis.phila.gov/gisdata/ODP/PWD_Parcels.zip' -L -o data/PWD_Parcels.zip
-curl 'https://phl.carto.com/api/v2/sql?q=SELECT+*+FROM+inlets&filename=inlets&format=shp&skipfields=cartodb_id' -L -o data/inlets.zip
+if ! [ -f data/STR_Centerline.zip ]; then
+    curl 'https://phila-gisdata.s3.amazonaws.com/ODP/STR_Centerline.zip' -L -o data/STR_Centerline.zip
+fi
+
+if ! [ -f data/PWD_Parcels.zip ]; then
+    curl 'http://gis.phila.gov/gisdata/ODP/PWD_Parcels.zip' -L -o data/PWD_Parcels.zip
+fi
+
+if ! [ -f data/inlets.zip ]; then
+    curl 'https://phl.carto.com/api/v2/sql?q=SELECT+*+FROM+inlets&filename=inlets&format=shp&skipfields=cartodb_id' -L -o data/inlets.zip
+fi
--- a/node_lambnik/scripts/fetch-data.sh
+++ b/node_lambnik/scripts/fetch-data.sh
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
 
 # Fetches development data sets from OpenDataPhilly
-if ! [ -f data/STR_Centerline.zip ]; then
-    curl 'https://phila-gisdata.s3.amazonaws.com/ODP/STR_Centerline.zip' -L -o data/STR_Centerline.zip
+if ! [ -f data/street_centerline.zip ]; then
+    curl 'https://phila-gisdata.s3.amazonaws.com/ODP/STR_Centerline.zip' -L -o data/street_centerline.zip
 fi
 
-if ! [ -f data/PWD_Parcels.zip ]; then
-    curl 'http://gis.phila.gov/gisdata/ODP/PWD_Parcels.zip' -L -o data/PWD_Parcels.zip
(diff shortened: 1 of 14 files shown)


node_lambnik/scripts/updatedb
```diff
@@ -150,6 +150,7 @@ check_postgis_install;
 
 for entry in data/*
 do
+	echo "Attempting to add $entry to database"
```
KlaasH: Trivial: Looks like another sneaky tab.

KlaasH (changes requested):
I think it would be nice for this stuff to be a little more self-contained and self-documenting.  So my idea would be to:
- Rename `updatedb` to `load_sample_data` and update the usage message to reflect what it does now--looks for files in `data/` and attempts to load them into the development DB, treating any `.zip` file as a zipped shapefile and any `.gz` file as a gzipped SQL dump.
- Add a `--download` option to download the default sample data files (i.e. what `fetch-data.sh` is doing here).

node_lambnik/scripts/fetch-data.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+# Fetches development data sets from OpenDataPhilly
+curl 'https://phila-gisdata.s3.amazonaws.com/ODP/STR_Centerline.zip' -L -o data/STR_Centerline.zip
+curl 'http://gis.phila.gov/gisdata/ODP/PWD_Parcels.zip' -L -o data/PWD_Parcels.zip
+curl 'https://phl.carto.com/api/v2/sql?q=SELECT+*+FROM+inlets&filename=inlets&format=shp&skipfields=cartodb_id' -L -o data/inlets.zip
```
KlaasH: It might be friendly to test whether the destination file exists and only download if it doesn't.  That would fit in well with the behavior of the DB import, which can be run repeatedly because it checks and moves on if its work is already done.

node_lambnik/src/demo/datatype-demo.html
```diff
@@ -16,14 +16,14 @@ <h3>Datatype Demo</h3>
     <div id="map" style="width: 800px; height: 800px"></div>
     <div>
         <form id="layer-selector">
-            <input type="radio" name="layer" id="pwd_inlets" value="pwd_inlets">
-            <label for="pwd_inlets">pwd_inlets (POINT/EPSG:4326)</label>
+            <input type="radio" name="layer" id="inlets" value="inlets">
```
KlaasH, line 19: Is there a reason for the "pwd_inlets"->"inlets" rename?  It's fine either way, just curious.

node_lambnik/scripts/fetch-data.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Fetches development data sets from OpenDataPhilly
+if ! [ -f data/STR_Centerline.zip ]; then
+    curl 'https://phila-gisdata.s3.amazonaws.com/ODP/STR_Centerline.zip' -L -o data/STR_Centerline.zip
```
KlaasH: I'm getting this when I run `update` a second time:
```
Attempting to add data/STR_Centerline.zip to database                                                                     
Loading STR_Centerline                                                                                                    
Loading data/STR_Centerline.zip                                                                                           
Attempting to extract data/STR_Centerline.zip                                                                             
Archive:  data/STR_Centerline.zip                                                                                         
  inflating: data/STR_Centerline/Street_Centerline.cpg                                                                    
  inflating: data/STR_Centerline/Street_Centerline.dbf                                                                    
  inflating: data/STR_Centerline/Street_Centerline.prj                                                                    
  inflating: data/STR_Centerline/Street_Centerline.sbn                                                                    
  inflating: data/STR_Centerline/Street_Centerline.sbx                                                                    
  inflating: data/STR_Centerline/Street_Centerline.shp                                                                    
  inflating: data/STR_Centerline/Street_Centerline.shp.xml                                                                
  inflating: data/STR_Centerline/Street_Centerline.shx                                                                    
Field length is an FTDouble with width 19 and precision 8                                                                 
Field shape_len is an FTDouble with width 19 and precision 11                                                             
Shapefile type: Arc
Postgis type: MULTILINESTRING[2]
SET
SET
BEGIN
ERROR:  relation "street_centerline" already exists
ERROR:  current transaction is aborted, commands ignored until end of transaction block
ERROR:  current transaction is aborted, commands ignored until end of transaction block
ERROR:  current transaction is aborted, commands ignored until end of transaction block
```
(and so on--many copies of that error)

The check for table existence is based on the filename, so they need to match (which is another thing to note in the usage message about how this actually finds and loads data).

KlaasH (changes requested):

node_lambnik/src/demo/datatype-demo.html
mattdelsordo: That's the name of the file on OpenDataPhilly, since we're downloading all of the data at update time now I thought it might be a good idea to keep the file names the same across sources. Do you think it'd be unnecessarily confusing?

node_lambnik/scripts/fetch-data.sh
mattdelsordo: I'll go in to the download script and make sure all of the downloaded filenames match the eventual table names, and add some documentation about adding data sets in this way.

node_lambnik/src/demo/datatype-demo.html
KlaasH: No, it just seemed like an intentional shift and I wasn't sure what the intention was, but it's fine either way.

mattdelsordo: Oh, now that I look, I think the other aspect of the change was that downloading the data straight from OpenDataPhilly leads to `inlets` being the name of the table in the database, and I thought it would be easier to update references than to remember to modify the table info.

node_lambnik/scripts/fetch-data.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Fetches development data sets from OpenDataPhilly
+if ! [ -f data/STR_Centerline.zip ]; then
```
KlaasH: The filename matches the table name now, but this check is still looking for the old filename (so it re-downloads every time).

KlaasH (approved):
Looks good.  It's doing the right thing now (not repeating steps unnecessarily, not throwing errors) when I run `update` repeatedly, except for the one download noted below.  :+1: with that fixup.

PR #74, opened 2018-07-23
Feature/vector tiles

### LINKED ISSUE #57: Vector tile processing initial demo

mattdelsordo:
## Overview

This PR enables serving vector tiles at the endpoint `/vector/{z}/{x}/{y}` as gzipped files that can be read by Mapbox GL JS. Layer filtering via query string is supported for consistency but can also be done client-side through Mapbox's library.

### Demo
vector styling example

## Testing Instructions

 * Start the server with `./scripts/update`, `./scripts/server`, etc.
 * Open up `demo/vector-tile-demo-mapbox.html`. Layers can be selected through the drop-down menu.
 * The styles for the currently viewed layer is printed in the text area. Change on of the colors/widths and click "Set Style". The changed should happen instantly as they are handled on the client rather than rendered on the server.

Resolves #57

### DIFF
--- a/node_lambnik/src/demo/vector-tile-demo-mapbox.html
+++ b/node_lambnik/src/demo/vector-tile-demo-mapbox.html
@@ -1,0 +1,140 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset="utf-8">
+    <title>Datatype Demo</title>
+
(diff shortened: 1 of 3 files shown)


KlaasH (approved):
:+1: This looks great!  Some cool new tricks in this demo.

PR #79, opened 2018-07-26
Feature/improve documentation

### LINKED ISSUE #48: Draft documentation for how to add a node-lambnik tiler to a project
Notes:
- It should ultimately be a Markdown file, but Google Docs is way nicer for drafting and collaborative document editing than GitHub/Markdown, so the working document should be a Google doc.  Which also means we should focus on text and do only very basic formatting for the initial pass.
- We should aim initially for a comprehensible and followable draft that covers the process from beginning to end.  So not spending extra time on streamlining any particular step until we've got all the steps down and can use the perspective gained from having the whole thing written out to help decide what steps need to be improved or streamlined.
- BUT its possible/likely that in writing up a basic workable draft we'll discover elements that we missed or that are implemented but in a way that's too unwieldy to even write down, in which case we should open issues for those and sort them out before documenting those parts of the process.

### LINKED ISSUE #53: Document required AWS permissions in README
We should add a section to the README laying out what AWS permissions are required for deploying node-lambnik.  The ones we've got enabled currently are:
- CloudFrontFullAccess
- AmazonAPIGatewayAdministrator
- AmazonRDSFullAccess
- AmazonS3FullAccess
- AWSLambdaFullAccess
- An inline policy to grant permissions to let Claudia.js create the role it needs without granting full IAM access:
   ```json
   {
       "Version": "2012-10-17",
       "Statement": [
           {
               "Effect": "Allow",
               "Action": [
                   "iam:DeleteRolePolicy",
                   "iam:DeleteRole",
                   "iam:CreateRole",
                   "iam:AttachRolePolicy",
                   "iam:PutRolePolicy"
               ],
               "Resource": "arn:aws:iam::*:role/tilegarden*"
           }
       ]
   }
   ```

### LINKED ISSUE #66: Add contents of Google Drive readme to git readme

mattdelsordo:
## Overview

This PR adds the content of the README that was being added to on Google Drive to the repository, spruces it up a bit, and links to it from the demo webpages. The navigation on the demo webpages has also been improved to accommodate this.

## Testing Instructions

 * [Read the new README.md here.](https://github.com/azavea/tilegarden/blob/feature/improve-documentation/README.md)
 * You have 50 minutes to read the passage and write an essay in response to the prompt provided inside this booklet.
   1. Do not write your essay in this booklet. Only what you write on the lined pages of your answer booklet will be evaluated.
   2. An off-topic essay will not be evaluated.
 * Your essay must be written on the lines provided in your answer booklet; except for the planning page of the answer booklet, you will receive no other paper on which to write. You will have enough space if you write on every line, avoid wide margins, and keep your handwriting to a reasonable size. Remember that people who are not familiar with your handwriting will read what you write. Try to write or print so that what you are writing is legible to those readers.

Resolves #48, resolves #53, resolves #66


README.md
```diff
@@ -1,26 +1,99 @@
...
+     * [UTF Grids](#utf-grids)
+     * [Vector Tiles](#vector-tiles)
+     * [Debugging](#debugging)
+   * [Deployment to AWS](#deployment-to-aws)
+     * [Additional Configuration Options](#additional-configuration-options)
+     * [Required AWS Permissions](#required-aws-permissions)
+ * [Helpful Links](#helpful-links)
+
+## About
+Tilegarden is a serverless tile-rendering tool using [Mapnik](http://mapnik.org/), built for AWS Lambda. Serve custom-generated map tiles without having to worry about server maintenance, scaling, or paying for resources that aren't being accessed.
```
KlaasH, line 18: Maybe mention PostGIS here, too.  The expectations built in around what the datasource is seem like information on the same level as this--helpful to know up front so you can understand the nature of the thing.

mattdelsordo: Are there actually expectations about the datasource? I'm pretty it'll support any datasource Mapnik does, I've tested it on local shapefiles and it's worked fine.

mattdelsordo: Although now that I think about it, to use something like local shapefiles you would probably have to store them in S3, and grabbing stuff from S3 isn't something we've implemented yet. Also, tiling and UTF grids sort of hinge on PostGIS being the datasource, so maybe I _should_ add that in.

README.md
```diff
@@ -1,26 +1,102 @@
...
+   * Make sure the name of the file matches the name of the sql table that contains the data.
+ * Configure your map:
+   * Open the file `src/tiler/src/config/map-config.mml`. This is where you specify your map settings. [A full specification for Carto’s .mml format can be found here.](https://cartocss.readthedocs.io/en/latest/mml.html)
+   * Of particular note is the “Layer” property, which specifies the layers of your map (as an array). Odds are you won’t have to change the target srs (at the top of the file), but make sure that the srs of each layer is specified properly.
+ * Create a copy of `env-template` named `.env`. This contains production-specific configuration options and doesn’t need to be filled out right now (but must exist in order to run the development server).
+ * Run `./scripts/update` to create your Docker containers and populate the development database. The optional flag `--download` will download the data sets used for our demos.
+ * Run `./scripts/server` to start the development server. It will be available at localhost:9001 (not 3000!), your tiles can be found at `/tile/{z}/{x}/{y}.png`.
```
mmcfarland: The `layers` query string parameter is only mentioned in the `Filter` section below.  It might be good to tie in that concept with the `Configure your map` section above, so one knows how to limit which specified layers are rendered.

demo/filter.html
```diff
@@ -4,6 +4,9 @@
     <meta charset="utf-8" />
     <title>Tilegarden</title>
     <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.2/css/bootstrap.min.css" integrity="sha384-Smlep5jCw/wG7hdkwQ/Z5nLIefveQRIY9nfy6xoR1uRYBtpZgI6339F5dgvm/e9B" crossorigin="anonymous">
+    <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha384-q8i/X+965DzO0rT7abK41JStQIAqVgRVzpbzo5smXKp4YfRvH+8abtTE1Pi6jizo" crossorigin="anonymous"></script>
```
mmcfarland, line 7: I'm sure no one is excited at the prospect, but these demo pages are getting more comprehensive and less DRY.  It could make making changes to shared parts of them (navigation, dependencies, etc) challenging or error prone.  Since it's likely required a build step to template these, I'm not suggesting we do it in this PR, but we may either want a future issue or at least keep an eye on how frequently they get edited and make a call down the road.

README.md
```diff
@@ -1,26 +1,102 @@
...
- * To publish, make sure you have specified valid AWS credentials and have listed the proper database credentials in a `.env` file in the root of the project. The format for this file can be copied from `env-template`. You can then run `./scripts/publish` to publish to AWS Lambda.
-    * For subsequent deployments of the same lambdas, leave off the `--new` tag.
-    * **NOTE:** Only `LAMBDA_TIMEOUT` and `LAMBDA_MEMORY` are changed on "update" deployments, all other settings must be changed by logging in to the Lambda Management Console or by deploying a new project.
- * API Gateway has trouble serving images, so you need to configure some sort of proxy (e.g. using a CloudFront distribution) that sets an `Accept:image/png` header on all lambda requests.
+### Local Development
+Dependencies: docker, docker-compose
+ * Clone the Tilegarden repository to your machine.
+ * Add any development data:
+   * Add your geospatial data to the `data/` directory. .zipped shapefiles and gzipped sql dumps will be loaded in to the development database automatically.
+   * Make sure the name of the file matches the name of the sql table that contains the data.
```
KlaasH: capitalize SQL

README.md
KlaasH: If you add the protocol, it will make the localhost URL a link, i.e. http://localhost:9001, which seems friendly.

KlaasH: Either added to this line or as another bullet, I think it makes sense to say something like "If you loaded the demo data sets, the HTML files in the `demo/` directory should show working demos when opened in a browser."  That's the actual payoff of the development setup--the point where you actually have something interesting to show that the project is spun up--so it makes sense to highlight it.

README.md
```diff
@@ -1,26 +1,102 @@
...
- * A new window will open up. Type `ctrl+p` to open a search bar that lets you navigate to your source code. **The transpiled code in `tiler/bin/` is what is actually being tracked by the debugger,**  not the source code in `tiler/src/`.
+#### Datasources
+Tilegarden supports the use of any geospatial data source that Mapnik/Carto does (shapefile, postgis, pgraster, raster). However, bear in mind that only PostGIS data sources support custom queries, and are thus the only ones that allow you to perform additional filtering on your data (see [Filters](#filters)). Also, attempting to bundle large local data files into your Tilegarden deployment could lead to rejection by AWS Lambda due to size restrictions.
+
+#### Map Styles
+[Map styles are specified in CartoCSS, the specification for which can be found here.](https://cartocss.readthedocs.io/en/latest/) The default stylesheet is located at `src/tiler/src/config/style.mss`. Multiple stylesheets can be used by adding them to the parameter “Stylesheet” in `src/tiler/src/config/map-config.mml`. Each .mss “class” refers to a map layer (its “id” property), specified in `map-config.mml`.
```
KlaasH: I would linkify only the second half of the first sentence.  I was trying to figure out what I expected the link to be pointing to until I got past the comma.

README.md
```diff
@@ -1,26 +1,102 @@
...
-    * **NOTE:** Only `LAMBDA_TIMEOUT` and `LAMBDA_MEMORY` are changed on "update" deployments, all other settings must be changed by logging in to the Lambda Management Console or by deploying a new project.
- * API Gateway has trouble serving images, so you need to configure some sort of proxy (e.g. using a CloudFront distribution) that sets an `Accept:image/png` header on all lambda requests.
+### Local Development
+Dependencies: docker, docker-compose
+ * Clone the Tilegarden repository to your machine.
+ * Add any development data:
+   * Add your geospatial data to the `data/` directory. .zipped shapefiles and gzipped sql dumps will be loaded in to the development database automatically.
+   * Make sure the name of the file matches the name of the sql table that contains the data.
+ * Configure your map:
+   * Open the file `src/tiler/src/config/map-config.mml`. This is where you specify your map settings. [A full specification for Carto’s .mml format can be found here.](https://cartocss.readthedocs.io/en/latest/mml.html)
```
KlaasH: I think you an make filenames into relative links, like this: [`src/tiler/src/config/map-config.mml`](src/tiler/src/config/map-config.mml)
That won't work in this comment because the link is relative to this PR, but in the actual README I believe it will link to the file relative to whatever branch you're reading the README on.
It might not be necessary in all cases, but I find it nice to have file references linked so I can quickly see what we're talking about while I'm reading about it.

README.md
```diff
@@ -1,26 +1,102 @@
...
+
+#### Map Styles
+[Map styles are specified in CartoCSS, the specification for which can be found here.](https://cartocss.readthedocs.io/en/latest/) The default stylesheet is located at `src/tiler/src/config/style.mss`. Multiple stylesheets can be used by adding them to the parameter “Stylesheet” in `src/tiler/src/config/map-config.mml`. Each .mss “class” refers to a map layer (its “id” property), specified in `map-config.mml`. 
+
+#### Filters
+Filters are specified by altering the query of one of your map’s layers in `src/tiler/src/config/style.mss`, and can then be fetched through the API. To create a filtered layer, modify the “table” property of the layer to have the format `(QUERY) as VARIABLE`. 
+   * Make sure each layer has unique “id” and “name” values.
+   * Make sure to include the geom column in this query, along with whatever columns you are filtering by.
```
KlaasH: This should either say "the `geom` column" or "the geometry column".  Since the `geometry_field` parameter in the `Datasource` MML spec lets you pick your column, probably the latter is better.

README.md
```diff
@@ -1,26 +1,102 @@
...
+   * *Example:* `(SELECT geom,homes FROM table_name WHERE homes=’big’ ) as q` would create a filtered layer for all points that have a “homes” value of “big” in your database.
```
KlaasH: Much as I like doing initial drafting in Google Docs, it has its perils--these tilty quotes should be replaced with basic ones, at least in the SQL example.

README.md
```diff
@@ -1,26 +1,102 @@
...
+
+#### Map Styles
+[Map styles are specified in CartoCSS, the specification for which can be found here.](https://cartocss.readthedocs.io/en/latest/) The default stylesheet is located at `src/tiler/src/config/style.mss`. Multiple stylesheets can be used by adding them to the parameter “Stylesheet” in `src/tiler/src/config/map-config.mml`. Each .mss “class” refers to a map layer (its “id” property), specified in `map-config.mml`. 
+
+#### Filters
+Filters are specified by altering the query of one of your map’s layers in `src/tiler/src/config/style.mss`, and can then be fetched through the API. To create a filtered layer, modify the “table” property of the layer to have the format `(QUERY) as VARIABLE`. 
+   * Make sure each layer has unique “id” and “name” values.
```
KlaasH: Is `name` required?  The demo ones are all identical to the `id`, so leaving it off (in the file and the instructions) would be preferable if it's allowed.

README.md
```diff
@@ -1,26 +1,102 @@
...
+Vector tiles of your data can be generated at the endpoint `/vector/{z}/{x}/{y}`. Filtering by layer works via query string but can also be handled client side, as can styling. [See here for more info about the Mapbox Vector Tile specification.](https://www.mapbox.com/vector-tiles/)
+
+#### Debugging
+The local development server exposes a Node.js process WebSocket that can be attached to your IDE of choice to step through and debug your code. Here are instructions on how to do so using Google Chrome's Dev Tools:
+ * Start the development server with `./scripts/server`.
+ * Open Google Chrome and navigate to `about:inspect`.
+ * There should be an option listed as something along the lines of "Target (v8.10.0)" with the Node.js logo and a path like `file:///home/tiler/bin`. Click "inspect", underneath.
+ * A new window will open up. Enter `ctrl+p` to open a search bar that lets you navigate to your source code. _The transpiled code in `src/tiler/bin/` is what is actually being tracked by the debugger,_  not the source code in `src/tiler/src/`.
+
+### Deployment to AWS
```
KlaasH, line 65: I think a bit of intro would be helpful here before jumping into the steps.  I was thinking something like "A deployed Tilegarden endpoint consists of 1) a Lambda function to generate tiles, 2) an API Gateway endpoint to interpret URL parameters and route to the correct entrypoint in the Lambda function and 3) a CloudFront distribution to work around a limitation in API Gateway's image handling."

I see that you've already got a component rundown like this below at the `./scripts/publish` step, but I think it would be better to put it up front and have the `publish` section refer to it.

README.md
```diff
@@ -1,26 +1,102 @@
...
+   * `LAMBDA_REGION`: the AWS region to which you want your lambda functions deployed.
+   * `LAMBDA_ROLE`: the name/ARN of the IAM role to give your lambda. Leave this blank to have one created for you automatically.
+   * `LAMBDA_SUBNETS` and `LAMBDA_SECURITY_GROUPS`: only required if you need to connect to other AWS resources (such as an RDS instance), in which case these should match the values that those resources have.
+   * [See below for more options.](#Additional-Configuration-Options)
+ * [See below to make sure your AWS profile has the requisite permissions for automated deployment.](#Required-AWS-Permissions)
+ * Run `./scripts/publish --new`. This will:
+   1. Create a new AWS Lambda function with your specified settings
+   2. Create an API Gateway from which to serve your content
+   3. Create a CloudFront distribution to act as a proxy to circumvent certain API Gateway limitations.
+ * Updates to an existing project can be deployed by running `./scripts/publish` (no “--new”).
```
KlaasH: It seems worth noting here what parameters can be changed by republishing and what ones require that you `destroy` the `publish --new` (unless I'm behind the times and republishing works for all configurable settings now).

README.md
```diff
@@ -1,26 +1,102 @@
...
+ * All deployed resources can be removed by running `./scripts/destroy`.
+
+#### Additional Configuration Options
+ * `USER`: Left unspecified so that the code knows who the local user is from inside docker containers. Gets tacked on to the end of the project name at deployment.
+ * `LAMBDA_TIMEOUT`: Time (in seconds) before your lambda function times out (maximum 300).
+ * `LAMBDA_MEMORY`: Amount of memory (in MB) to allocate per function. Must be a multiple of 64, minimum 128, maximum 3008.
+
+#### Required AWS Permissions
+For automated deployment, the AWS profile you are deploying as has to have at least the following permissions:
```
KlaasH: "deploying as has to have" seems like more words than we need.  Maybe change the whole sentence to "The AWS profile used for deployment must have at least the following policies and permissions:"

(the added "policies and" is because the first three aren't permissions, exactly, but rather standard groups of permissions that they call "managed policies")

README.md
```diff
@@ -1,26 +1,102 @@
...
+ * `CloudFrontFullAccess`
+ * `AmazonAPIGatewayAdministrator`
+ * `AWSLambdaFullAccess`
+ * If you don’t specify a pre-existing execution role, at least the following IAM permissions
+   * `iam:DeleteRolePolicy`
+   * `iam:DeleteRole`
+   * `iam:CreateRole`
+   * `iam:AttachRolePolicy`
+   * `iam:PutrolePolicy`
```
KlaasH: Capitalize "role", i.e. `PutRolePolicy`

demo/filter.html
KlaasH: Yeah, definitely a lot of repetition happening.  I think making an issue makes sense--having demos of the range of capabilities of the thing is useful, so I think it's more likely that they get ongoing attention/edits than that they're done and settled.

KlaasH (changes requested):
Some various content and copy suggestions below, but this looks good.

demo/filter.html
mattdelsordo: I agree - I was considering this when putting the demos together, but since the only templating engine I'm super familiar with is React (which is very overkill for this use case, in my opinion) I didn't think the time cost of familiarizing myself with something like Pug + setting that up + re-implementing everything with templates would be worth it. I'll add an issue for this and revisit it if the pages get more numerous/complex.

README.md
mattdelsordo: That it a good question.
EDIT: [The documentation](https://cartocss.readthedocs.io/en/latest/mml.html#name) simply says "A name for the stylesheet", which doesn't explain a lot, but leaving it out doesn't seem to have any real effect.

README.md
```diff
@@ -1,26 +1,105 @@
...
+ * Create a copy of [`env-template`](env-template) named `.env`. This contains production-specific configuration options and doesn't need to be filled out right now (but must exist in order to run the development server).
+ * Run `./scripts/update` to create your Docker containers and populate the development database. The optional flag `--download` will download the data sets used for our demos.
+ * Run `./scripts/server` to start the development server. It will be available at [localhost:3000](http://localhost:3000/), your tiles can be found at `/tile/{z}/{x}/{y}.png`.
+ * The local development server exposes a node.js debugger, which can be attached to by Chrome DevTools or your IDE of choice [(see below)](#Debugging).
+   * (Optional) If you downloaded the example data using `./scripts/update --download`, opening `index.html`(index.html) (or any of the pages in `demo/`) should show you working demos.
```
KlaasH: Looks like `index.html` was intended to be in square brackets so it would linkify.

KlaasH (approved):
:+1: Looks great!

PR #84, opened 2018-07-31
Feature/remove nginx container

### LINKED ISSUE #81: Remove the proxy container from docker-compose
After fixing claudia-local-api, the dev server has no issue just serving binary data, so having a proxy container to add the correct headers/mimic cloud front isn't doing anything.

mattdelsordo:
## Overview

After the binary content issues with claudia-local-api were fixed, it turned out that using nginx in development to mimic the behavior of the production CloudFront distribution was unnecessary. claudia-local-api uses express under the hood, which lacks API Gateway's binary content serving issues. This PR removes the nginx docker container from the project and switches the port of the development server from 9001 to 3000, to cut down on confusion from the message that the server prints on startup.

## Testing Instructions

 * Navigate to http://localhost:3000. The expected landing page should be there, instead of at port 9001.

Resolves #81


mmcfarland (approved):
Tiles are served correctly, even without the `accept` header specified.  May be good to keep a note somewhere that the API Gateway does have this constraint, even if local dev does not.

PR #85, opened 2018-07-31
Upgrade claudia-local-api to a version that allows for output reduction

### LINKED ISSUE #83: Reduce level of console output
By default, the local development server prints out the contents of all requests/responses to the server. This is helpful in general but not when each response body contains a utf-encoded image file, that tends to fill up the entire console window and block other information.

mattdelsordo:
## Overview

I've added another PR to claudia-local-api which implements an `--abbrev` flag which lets the user specify an amount to trim `loggedOuput.body` to. This makes it easier to read the development server logs and noticeably speeds up the development server since it isn't spending so much time printing output.

See the PR I submitted here: https://github.com/suddi/claudia-local-api/pull/20, https://github.com/mattdelsordo/claudia-local-api/pull/4

### Demo

What output tends to look like now with `--abbrev 200`. Normally this body string is over 10k characters long (a 256x256 .png).
```
 19:07:13.775Z  INFO claudia-local-api:
     {
         "statusCode": 200,
         "headers": {
             "Content-Type": "image/png",
             "Access-Control-Allow-Origin": "*",
             "Access-Control-Allow-Headers": "Content-Type,Authorization,X-Amz-Date,X-Api-Key,X-Amz-Security-Token",
             "Access-Control-Allow-Methods": "GET,OPTIONS",
             "Access-Control-Allow-Credentials": "true",
             "Access-Control-Max-Age": "0"
        },
         "body": "iVBORw0KGgoAAAANSUhEUgAAAQAAAAEACAYAAABccqhmAABG6ElEQVR4nO19B5iUVbL26zDMwAw55wwSJElGlJwkqgTFiAEDioqo[...]+BIog6lSvEDcHIQ8hIMEqIcGrqcTTJHyXe4l7UBJ+z3O00+cjYWh7sbi1YT0HDB4f8DG3dx5Umi99kAAAAASUVORK5CYII=",
         "isBase64Encoded": true
     }
```

## Testing Instructions

 * Start the development server with `./scripts/update` and `./scripts/server`.
 * Open up one of the demo pages locally. The resulting logs should be a lot easier on the eyes, and on the server.

Resolves #83

### DIFF
--- a/src/tiler/package.json
+++ b/src/tiler/package.json
@@ -29,7 +29,7 @@
     "destroy": "claudia destroy",
     "dev": "nodemon -e js,mss,json,mml,mss --ignore bin/ --exec yarn local",
     "lint": "eslint src",
-    "local": "yarn transpile && node --inspect=0.0.0.0:9229 node_modules/claudia-local-api/bin/claudia-local-api --api-module bin/api | bunyan -o short",
+    "local": "yarn transpile && node --inspect=0.0.0.0:9229 node_modules/claudia-local-api/bin/claudia-local-api --abbrev 300 --api-module bin/api | bunyan -o short",
     "parse-id": "jq -r '.api.id' claudia.json > .api-id",
     "test": "eslint src && jest --coverage",
     "transpile": "yarn build-xml && babel src -d bin --source-maps inline"
@@ -41,7 +41,7 @@
     "bunyan": "^1.8.12",
     "carto": "^1.0.1",
     "claudia": "^5.0.1",
-    "claudia-local-api": "https://github.com/mattdelsordo/claudia-local-api.git#feature/binary-response-support",
+    "claudia-local-api": "https://github.com/mattdelsordo/claudia-local-api.git",
     "eslint": "^4.19.1",
     "eslint-config-airbnb-base": "^12.1.0",
     "eslint-plugin-import": "^2.12.0",
--- a/src/tiler/yarn.lock
+++ b/src/tiler/yarn.lock
@@ -1197,16 +1197,14 @@ claudia-api-builder@^4.1.0:
   version "4.1.1"
   resolved "https://registry.yarnpkg.com/claudia-api-builder/-/claudia-api-builder-4.1.1.tgz#9198280c6f5ef19a5fceadf716a819f1d6fc5ce0"
 
-"claudia-local-api@https://github.com/mattdelsordo/claudia-local-api.git#feature/binary-response-support":
+"claudia-local-api@https://github.com/mattdelsordo/claudia-local-api.git":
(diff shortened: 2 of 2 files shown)


KlaasH:
This is failing to install for me, with a message that includes
> Couldn't find match for \"b33c2e2852ea1ca3b61948462462b8ee49fa8d1e\" 

That's the commit hash in the `yarn.lock` file and I don't see it in the branch, which makes me think maybe the branch was rebased and `yarn.lock` needs to be updated to match.

This is related to my next thought, which is that it makes more sense to merge these modifications into your fork of `claudia-local-api` and use the fork's master branch than to point the dependency at a branch, especially when that branch is ending up being a stack of different feature branches.

That would also mean there would be a place to discuss the details of the `claudia-local-api` changes.  I have a couple comments on that part of this PR, but no real place to put them, since they're not actually part of this PR.

I'd vote for making PRs on your fork for each of the branches you PRed upstream, then we can discuss and merge them there (but still keep the branches open for the purposes of the upstream PRs) and point this at `master`.

KlaasH (commented):
This is failing to install for me, with a message that includes
> Couldn't find match for \"b33c2e2852ea1ca3b61948462462b8ee49fa8d1e\" 

That's the commit hash in the `yarn.lock` file and I don't see it in the branch, which makes me think maybe the branch was rebased and `yarn.lock` needs to be updated to match.

This is related to my next thought, which is that it makes more sense to merge these modifications into your fork of `claudia-local-api` and use the fork's master branch than to point the dependency at a branch, especially when that branch is ending up being a stack of different feature branches.

That would also mean there would be a place to discuss the details of the `claudia-local-api` changes.  I have a couple comments on that part of this PR, but no real place to put them, since they're not actually part of this PR.

I'd vote for making PRs on your fork for each of the branches you PRed upstream, then we can discuss and merge them there (but still keep the branches open for the purposes of the upstream PRs) and point this at `master`.

mmcfarland:
> I'd vote for making PRs on your fork for each of the branches you PRed upstream

> use the fork's master branch than to point the dependency at a branch

Seconded on both, :+1:

mattdelsordo:
Ah yeah, I definitely rebased that branch to fix up some stuff for the CI tests after adding it to the project, not realizing that would happen. Does pointing to master get handled differently than other commits, as in, it's not stored in yarn.lock as a commit hash?

KlaasH:
> Does pointing to master get handled differently than other commits, as in, it's not stored in yarn.lock as a commit hash?

No, it still gets resolved to a commit hash.  But it would be much less likely to end up pointing at a nonexistent commit hash.

KlaasH (approved):
:+1:

PR #87, opened 2018-08-01
Feature/multiple config

### LINKED ISSUE #80: Allow for multiple configs that can be dynamically selected via query param
This way we don't have to have one monolithic .mml file specifying every conceivable configuration.

mattdelsordo:
## Overview
This PR is in two small parts:
1. The [`build-xml`](src/tiler/src/util/build-xml.js) utility script has been modified to automatically transpile all `.mml` files in the `src/tiler/src/config` directory. The script now takes input and output directories, rather than in/out files, which are set to `src/tiler/src/config` and `src/tiler/bin/config` in the build step.
2. The query string `config` can now be optionally set to the name of a configuration file (without the file extension) in order to dynamically load any `.xml` files in `src/tiler/bin/config`. It defaults to "map-config".

### Demo
`http://localhost:3000/tile/13/2385/3103.png?config=data-type-example`
image
`http://localhost:3000/tile/13/2385/3103.png?config=point-example`
image

## Testing Instructions

 * Spin up the development server with `./scripts/update` and `./scripts/server`.
 * Visit http://localhost:3000/tile/13/2385/3103.png. The server will default to using [`src/config/map-config.mml`](src/tiler/src/config/map-config.mml)'s configuration.
 * There are two other included configuration files, [`data-type-example.mml`](src/tiler/src/config/data-type-example.mml) and [`point-example.mml`](src/tiler/src/config/point-example.mml). Add those file names to http://localhost:3000/tile/13/2385/3103.png?config= to see what happens.

Resolves #80

### DIFF
--- a/src/tiler/src/api.js
+++ b/src/tiler/src/api.js
@@ -66,8 +66,9 @@ api.get(
         try {
             const { z, x, y } = processCoords(req)
             const layers = processLayers(req)
+            const { config } = req.queryString
 
-            return image(z, x, y, layers)
+            return image(z, x, y, layers, config)
                 .catch(e => messageTile(e.toString()))
         } catch (e) {
             return messageTile(e.toString())
@@ -85,8 +86,9 @@ api.get(
             const { z, x, y } = processCoords(req)
             const utfFields = processUTFQuery(req)
             const layers = processLayers(req)
+            const { config } = req.queryString
 
-            return grid(z, x, y, utfFields, layers)
+            return grid(z, x, y, utfFields, layers, config)
                 .catch(e => JSON.stringify(e))
         } catch (e) {
             return JSON.stringify(e)
@@ -100,8 +102,9 @@ api.get(
     (req) => {
         const { z, x, y } = processCoords(req)
         const layers = processLayers(req)
+        const { config } = req.queryString
 
-        return vectorTile(z, x, y, layers)
+        return vectorTile(z, x, y, layers, config)
     },
     VECTOR_RESPONSE,
 )
--- a/README.md
+++ b/README.md
@@ -4,6 +4,7 @@
  * [About](#About)
  * [Usage](#Usage)
    * [Local Development](#local-development)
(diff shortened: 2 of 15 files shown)


src/tiler/src/api.js
```diff
@@ -66,8 +66,9 @@ api.get(
         try {
             const { z, x, y } = processCoords(req)
             const layers = processLayers(req)
+            const { config } = req.queryString
```
KlaasH, line 69: Destructuring assignment is cool, but for only one variable it seems like it the [Explicit is better than implicit]() principle kicks in and `const config = req.queryString.config` would be preferable.

src/tiler/src/api.js
```diff
@@ -85,8 +86,9 @@ api.get(
             const { z, x, y } = processCoords(req)
             const utfFields = processUTFQuery(req)
             const layers = processLayers(req)
+            const { config } = req.queryString
```
KlaasH, line 89: Now that there are three things repeated three times I'm tempted to call for them to get bundled up in another helper function, but that's probably excessive.  Especially since there's also a 4th thing here, so clearly there's some variation in what's needed.  Though speaking of which, if the three common lines do stay repeated, it might be nice to move the `utfFields` line to the end to make it clearer what's the same and what's different between the methods.

KlaasH (approved):
A few style notes below, but this looks good :+1: 

The more MML files there are with big `Layer` config sections, the more I think about how we might be able to specify that stuff in a more succinct and reusable way.  But that's a matter for a separate issue, and probably not actually high-priority.

src/tiler/src/api.js
mattdelsordo: I agree, 100%. My reasoning for doing it like this was that a) eslint was yelling at me to do it like this and b) writing the assignment like this was a retroactive change to make the transition from this PR to the next commit (in the s3-config branch) more of an expansion on the same pattern, rather than a complete rewrite.

PR #90, opened 2018-08-02
Feature/s3 config

### LINKED ISSUE #26: Support loading map/style config from S3
Add an option to load layer and styling config from a file on S3.  We can create the bucket and upload the file manually for now.  It should use the configured AWS profile so that it’ll work with a private bucket.

mattdelsordo:
## Overview

This PR allows you to load configuration files from an S3 bucket rather than only locally. Adding a query string such as `?config=config-file.xml&s3bucket=tilegarden-bucket` will get you the config file with the S3 key "config-file.xml" that is stored in the bucket "tilegarden-bucket", as long as you deployed the lambda function using AWS credentials that have S3 read/write access. It will also work in local development, provided such AWS credentials are specified in your `.env`. 

In pursuit of this, a new script `build-xml` has been added to build a directory's worth of MML and MSS into a directory of XML without having to spin up the entire development apparatus and dig for the compiled files in `tiler/bin/`. It has several options that look like the following:
```
Usage: build-xml -i <path> -o <path>
Transpiles a directory's worth of MML+MSS files into Tilegarden-readable XML
Options:
    -h, --help          Display this help text
    -i, --in <path>     Relative path to input directory [src/tiler/src/config]
    -o, --out <path>    Relative path to output directory [transpiled-configs/]
    -b, --bucket <name> Uploads XML to the specified S3 bucket, instead of saving to disk (requires AWS CLI installed on your machine, uses credentials from .env)
    -d, --development   Build using development variables, instead of production ones
```
This script lead to a re-write of the whole XML-building pipeline. Most of the logic was relegated to bash scripts, and Carto was removed as a project dependency and instead installed as a global utility on the tiler container. 

### Demo
```
$ ./scripts/build-xml -d -b tilegarden-dev
Transpiling src/tiler/src/config//data-type-example.mml => /tmp/tmp.naNUwi4M32/data-type-example.xml
Starting tilegarden_database_1 ... done
Transpiling src/tiler/src/config//map-config.mml => /tmp/tmp.naNUwi4M32/map-config.xml
Starting tilegarden_database_1 ... done
Transpiling src/tiler/src/config//point-example.mml => /tmp/tmp.naNUwi4M32/point-example.xml
Starting tilegarden_database_1 ... done
Uploading files to s3://tilegarden-dev/...
upload: ../../../../tmp/tmp.naNUwi4M32/data-type-example.xml to s3://tilegarden-dev/data-type-example.xml
upload: ../../../../tmp/tmp.naNUwi4M32/map-config.xml to s3://tilegarden-dev/map-config.xml
upload: ../../../../tmp/tmp.naNUwi4M32/point-example.xml to s3://tilegarden-dev/point-example.xml
Cleaning up /tmp/tmp.naNUwi4M32...
```

`http://localhost:3000/tile/13/2385/3103.png?config=point-example.xml&s3bucket=tilegarden-dev`:
image

## Testing Instructions
 * Build the docker containers with `./scripts/update`.
 * Use `build-xml` to create some XML files and upload them to the S3 bucket of your choice. 
   * _Ex._ `./scripts/build-xml --development --bucket tilegarden-dev` to build XML files with development database connection info from the default input directory at `src/tiller/src/config` and push them to the S3 bucket "tilegarden-dev" that I've already created.
   * If you use the "tilegarden-dev" bucket, there might already be config files in there, so it might be a good idea to modify/rename the MSS/MML file you're testing with so you can note the difference.
 * Navigate to `http://localhost:3000/tile/13/2385/3103.png?config=<config file name>&s3bucket=<S3 bucket name>`. Switch between different config files to see what happens.

Resolves #26


scripts/build-xml
```diff
@@ -0,0 +1,106 @@
...
+
+        # Load AWS profile settings from env, if possible
+        if [[ -f .env ]]; then
+            set -a
+            source ./.env
+            set +a
+        fi
+
+        for config in $(echo "${out_dir}/*"); do
+            aws s3 cp "${config}" ${s3_url} --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers
```
KlaasH: Was the Lambda function not able to read the config files without this?  This appears to be granting global (including public) read access, which is not what we want.  I was able to load https://s3.amazonaws.com/tilegarden-dev/map-config.xml in an anonymous browser window.

scripts/build-xml
```diff
@@ -0,0 +1,106 @@
...
+        if [ "$ext" == "mml" ]; then
+            # get output path
+            filename="${file##*/}"
+            base="${filename%%.*}"
+            outPath="${out_dir}/${base}.xml"
+
+            mkdir -p "${out_dir}"
+
+            echo "Transpiling ${file} => ${outPath}"
+            docker-compose run -e "NODE_ENV=${node_env}" -v "$PWD/${in_dir}:/home/tiler/configs" tiler ./scripts/build-xml.sh "/home/tiler/configs/${base}.mml" "$(< ${file})" > ${outPath}
```
KlaasH: Are there differences between this section and `build-all-xml.sh` besides the output directory?  Would it work to map that to a volume as well and make this command call `build-all-xml.sh` instead of repeating it?

scripts/build-xml
mattdelsordo: To be honest I forgot why I did this, but thanks for pointing it out because for sure these files shouldn't be publicly accessible.

scripts/build-xml
mattdelsordo: When docker creates files in a volume, it doesn't necessarily grant read/write access to the executor of the initial script, so I tried to circumvent this by echoing the output from `build-xml.sh` into the non-docker context.

KlaasH: Oh right, the permissions.  Would be nice to not have to keep working around those.  But yeah, this makes sense.

scripts/build-xml
```diff
@@ -0,0 +1,106 @@
...
+            exit 1
+        fi
+
+        s3_url="s3://${s3_bucket}/"
+        echo "Uploading files to ${s3_url}..."
+
+        # Load AWS profile settings from env, if possible
+        if [[ -f .env ]]; then
+            set -a
+            source ./.env
```
KlaasH: This fails with `./.env: line 3: USER: command not found`
The bare variable name tells docker-compose to use the existing value but doesn't mean anything to bash.  What's in your `.env` right now?  Changing mine to `USER=$USER` worked for this, but I didn't test that it works the same for deployment.

mattdelsordo: It looks like I added `USER=mdelsordo` to my `.env` at some point and forgot to make that change in `env-template` as well.

mattdelsordo: Update: yep, it does work to deploy.

README.md
```diff
@@ -35,10 +35,14 @@ Dependencies: docker, docker-compose
...
  * _Example_: if you have a configuration file named `my-good-map.mml`, you can tell Tilegarden to use it with the endpoint `/tile/{z}/{x}/{y}.png?config=my-good-map`
  
+ Configuration files can also be optionally loaded from an AWS S3 bucket, which allows you to add/remove/replace available maps without having to redeploy your entire project. To do so, add `&s3bucket=<bucket-name>` to your query string, and treat the `config` parameter as the desired config file's key. You can use the `build-xml` script to generate config files for this purpose without having to spin up the entire development environment, run `./scripts/build-xml --help` for more info.
+  * Accessing config files stored in S3 requires you to have deployed your Tilegarden instance using an AWS profile with read/write permissions for the desired S3 bucket. These permissions are passed on to the AWS Lambda function.
+  * `build-xml` has a `-b/--bucket <name>` flag which automatically uploads built files to S3 instead of saving them to disk. This requires you to have the AWS CLI installed on your machine, and uses the AWS credentials you've specified in `.env`.
```
KlaasH: the flag is `--s3_bucket` (not `--bucket`)

scripts/build-xml
```diff
@@ -0,0 +1,106 @@
...
+node_env='production'
+
+function usage() {
+    echo -n "Usage: $(basename "${0}") -i <path> -o <path>
+Transpiles a directory's worth of MML+MSS files into Tilegarden-readable XML
+Options:
+    -h, --help          Display this help text
+    -i, --in <path>     Relative path to input directory [${in_dir}]
+    -o, --out <path>    Relative path to output directory [${out_dir}]
+    -b, --bucket <name> Uploads XML to the specified S3 bucket, instead of saving to disk (requires AWS CLI installed on your machine, uses credentials from .env)
```
KlaasH: the flag is `--s3_bucket` (not `--bucket`)

KlaasH (commented):
This is now working for me locally but on AWS I get > 2018-08-16T23:22:53.605Z 449cdb01-a1ab-11e8-a909-b9e57c251f33 { AccessDenied: Access Denied
> at Request.extractError (/var/task/node_modules/aws-sdk/lib/services/s3.js:577:35)
> at Request.callListeners (/var/task/node_modules/aws-sdk/lib/sequential_executor.js:105:20)
> at Request.emit (/var/task/node_modules/aws-sdk/lib/sequential_executor.js:77:10)
> at Request.emit (/var/task/node_modules/aws-sdk/lib/request.js:683:14)
> at Request.transition (/var/task/node_modules/aws-sdk/lib/request.js:22:10)
> at AcceptorStateMachine.runTo (/var/task/node_modules/aws-sdk/lib/state_machine.js:14:12)
> at /var/task/node_modules/aws-sdk/lib/state_machine.js:26:10
> at Request.<anonymous> (/var/task/node_modules/aws-sdk/lib/request.js:38:9)
> at Request.<anonymous> (/var/task/node_modules/aws-sdk/lib/request.js:685:12)
> at Request.callListeners (/var/task/node_modules/aws-sdk/lib/sequential_executor.js:115:18)
> message: 'Access Denied',
> code: 'AccessDenied',
> region: null,
> time: 2018-08-16T23:22:53.586Z,
> requestId: '5D376A7F5DB9D5C6',
> extendedRequestId: 'YuIARTwG3X8hM5wdlpM0JkV1GGe3xq+9vNZG94M6FLLom+EwKNL0kvmZ7ngaHi1/UpGB3f0MxjY=',
> cfId: undefined,
> statusCode: 403,
> retryable: false,
> retryDelay: 34.76702489574455 }

Which I don't understand, because I would have expected any resource within the account to be able to see the bucket, and I can certainly see the files with the AWS CLI using the same profile I'm using to deploy.

KlaasH (approved):
So yeah, looks like it's the role permissions.  Adding `AmazonS3ReadOnlyAccess` to my `tilegarden-dev-khoekema-executor` IAM role made using S3 config work remotely.

So I think we should do that manually for now, make an issue for integrating it somehow (we might need to use terraform to make a customized role), and merge this.

:+1:

PR #93, opened 2018-08-08
Feature/client side filtering

### LINKED ISSUE #89: Revisit client-side filtering
Windshaft does this somehow and it would be incredibly useful to have

mattdelsordo:
## Overview

This PR modifies the layer querying API to accept a JSON array rather than a comma-separated list of layer names, in order to allow for higher degrees of filtering to be done through the API. Filters applied via the API are templated around a layer's existing SQL query in the form of `SELECT * FROM <existing query> WHERE <column> <operator> <value>`. Multiple `WHERE` clauses can be strung together with `AND` and `OR`. The reasoning for these decisions is as follows:
 * Developers should be able to generate dynamic sets of tiles that vary with user input, but 
 * API users shouldn't have too much access to database information in order to prevent confidential information from leaking.
* Wrapping pre-existing queries allows developers to restrict the amount of information that is query-able.
 * It is safer to parse a JSON string than to invent a less-verbose data format and attempt to parse that using regexes, or something.

### Demo
filter-by-query-demo

### Notes

Caveats:
 * Using JSON makes the API a lot less human-readable/human-editable (but at lot easier for machines).
 * Changing the expected input for the `?layers` query string makes me think the other query strings that work using lists should expect as JSON arrays as well, although I don't want to fix what isn't broken.
 * Most browsers have a length limit on URLs (in the 1000s of chars) which could hypothetically cause issues for elaborate requests.

## Testing Instructions

 * Start the development server using `./scripts/update` and `./scripts/server`.
 * Open up `demo/client-filtering.html`.
 * Modify the JSON in order to modify the visible layer.
   * It may be helpful to open up the database with `./scripts/console database` and execute some queries to determine what data exists in there to see. `\d+ pwd_parcels;` and `SELECT DISTINCT owner2, count(*) FROM pwd_parcels GROUP BY owner2 ORDER BY count(*) DESC;` are what I looked at.

Resolves #89

### DIFF
--- a/demo/client-filter.html
+++ b/demo/client-filter.html
@@ -17,7 +17,7 @@
 
         document.addEventListener('DOMContentLoaded', function pageLoaded() {
             var map = L.map('map', {
-                center: [39.99, -75.15], zoom: 11
+                center: [39.97, -75.15], zoom: 11
             });
 
             // add basemap
@@ -37,28 +37,50 @@
 
             // add filtered layers
             function newLayer (filters) {
-                return L.tileLayer(TILESERVER_PATH + 'tile/{z}/{x}/{y}.png?config=point-example&layers=' + filters)
+                return L.tileLayer(TILESERVER_PATH + 'tile/{z}/{x}/{y}.png?config=data-type-example&layers=' + encodeURIComponent(JSON.stringify(filters)))
             }
 
             var EXAMPLE_PATH = 'https://yourtiles.cloudfront.net/tile/{z}/{x}/{y}.png?layers='
 
+            var INITIAL_QUERY = [
+                {
+                    name: 'pwd_parcels',
+                    mode: 'AND',
+                    filters: [
+                        {
+                            col: 'owner2',
+                            val: 'OF PHILADELPHIA',
+                        },
+                        {
+                            col: 'gross_area',
+                            op: '>',
+                            val: '3000'
+                        }
+                    ]
+                }
+            ]
+
             // keep track of active layers so
             var currentSelection;
-            document.getElementById("selector").addEventListener('mouseup', function selectorMouseUp() {
+            document.getElementById("btnUpdateQuery").addEventListener('click', function selectorMouseUp() {
                 // remove old layer
                 map.removeLayer(currentSelection);
 
-                // get new layer names from selector
-                var layerNames = Object.values(this.selectedOptions).map(function(v) {
-                    return v.value;
-                }).join(',');
-
-                currentSelection = newLayer(layerNames);
-                currentSelection.addTo(map);
-                document.getElementById('urlExample').innerText = EXAMPLE_PATH + layerNames;
+                // get new layer schema
+                var layerObj;
+                try {
+                    layerObj = JSON.parse(document.getElementById('querybox').value);
+                    currentSelection = newLayer(layerObj);
+                    currentSelection.addTo(map);
+                    document.getElementById('urlExample').innerText = EXAMPLE_PATH + JSON.stringify(layerObj);
+                } catch (e) {
+                    console.error(e);
+                }
             });
 
-            currentSelection = newLayer('PRIV,PARK');
+            currentSelection = newLayer(INITIAL_QUERY);
+            document.getElementById('urlExample').innerText = EXAMPLE_PATH + JSON.stringify(INITIAL_QUERY);
+            document.getElementById('querybox').value = JSON.stringify(INITIAL_QUERY, null, 2);
             map.addLayer(currentSelection);
         });
     </script>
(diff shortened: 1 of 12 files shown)


src/tiler/src/util/layer-filter.js
```diff
@@ -14,15 +14,91 @@ const parsePromise = xmlString => new Promise((resolve, reject) => {
...
+            case '<>':
+            case '!=':
+                /* eslint-disable-next-line prefer-destructuring */
+                op = filterObj.op
+                break
+            default:
+                op = '='
+                break
+            }
+        }
```
mmcfarland: It's not immediately clear that these switch blocks aren't really transforming the value (except maybe the case) to a sql-safe value, but it's really more of a white list.  Is the functionality preserved if you do something similar to:

```js
const whitelist = ['LIKE', 'IS', 'IS NOT', 'etc...'];
const modes = ['OR', 'AND'];

const op = whitelist.includes(filterObj.op.toUpperCase()) ? filterObj.op : '=';
const mode = modes.includes(filterObj.mode.toUpperCase()) ? filterOjb.mode : 'AND';
```

src/tiler/tests/layer-filter.test.js
```diff
@@ -1,7 +1,163 @@
...
+                {
+                    col: 'col',
+                    val: 'val'
+                }
+            ]
+        }
+        const existingQuery = 'table'
+        const expectedQuery = "(SELECT * FROM table WHERE col='val' AND col='val') as m"
+
+        expect(structureQuery(existingQuery, layer)).toBe(expectedQuery)
```
mmcfarland: I think this ought to be a failure condition, rather than a correct query that the author didn't expect (or write).

src/tiler/src/util/layer-filter.js
mmcfarland: After reviewing the tests, I'd amend my above sample to throw an exception (invalid syntax) if it failed the whitelist test, instead of defaulting to `=/AND`

mmcfarland (commented):
General approach looks good, had a question about the best way to handle unsupported queries.  Working through getting this branch running to test things out.

src/tiler/src/util/layer-filter.js
```diff
@@ -16,6 +16,65 @@ const parsePromise = xmlString => new Promise((resolve, reject) => {
...
+
+
+        query += `${mode ? ` ${mode}` : ''} ${filterObj.col}${op}'${filterObj.val}'`
```
mmcfarland: I'd need a little time to think of a good alternative, but there is a basic sql injection attack that is exposed with the main method of constructing the query here.  I was able to get a TRUNCATE statement to execute (though it failed, due to db permissions that may or may not be present in a different application):

```json
[
  {
    "name": "pwd_parcels",
    "mode": "AND",
    "filters": [
      {
        "col": "owner2",
        "val": "') as xxx; TRUNCATE pwd_parcels; --"
      }
    ]
  }
]
```

> Error: Postgis Plugin: ERROR:  cannot execute TRUNCATE TABLE in a read-only transaction

The attack was made easier to construct based on the error tiles which included the sql parsing errors. The postgis plugin for mapnik may always make this a read-only transaction, which would prevent drop/truncate/insert etc, but seems like an attacker with more motivation than me could figure out a way to abuse the privilege.  Is there a way to sanitize this input before concatenating strings?

mmcfarland (changes requested):
I think we need to revisit the way we construct the query from the user input, it appears problematic from a sql injection perspective.  See PR comment for details.

src/tiler/src/util/layer-filter.js
mattdelsordo: Those are good ideas, and definitely my intent with that bit.

src/tiler/src/util/layer-filter.js
mattdelsordo: I ditched the idea of sanitizing the inputs early on (when the filtering paradigm was still "just put the whole query in a url parameter) but forgot to re-introduce it now that the only variables in play are the column name and value! I'll get on that. Regarding the error tile issue: I've removed them in #95 and also filtered caught errors so that the Postgis ones never make it out of the back end.

demo/client-filter.html
```diff
@@ -0,0 +1,155 @@
...
+            // keep track of active layers so
+            var currentSelection;
+            document.getElementById("btnUpdateQuery").addEventListener('click', function selectorMouseUp() {
+                // remove old layer
+                map.removeLayer(currentSelection);
+
+                // get new layer schema
+                var layerObj;
+                try {
+                    layerObj = JSON.parse(document.getElementById('querybox').value);
```
KlaasH, line 72: I tried to do a `LIKE` query and it crashed because the `%` wildcards confused escaping.  Escaping them myself worked, i.e.:
```json
[
  {
    "name": "pwd_parcels",
    "mode": "AND",
    "filters": [
      {
        "col": "owner2",
        "val": "%25PHILA%25",
"op": "LIKE"
      },
      {
        "col": "gross_area",
        "op": "<",
        "val": "3000"
      }
    ]
  }
]
```
Not sure if it's worth the trouble to try to escape the `val` values here in the demo, but it might be worth noting that gotcha in the documentation.

mattdelsordo: It shouldn't be too big of a deal to add a catch for that case in. Are there any other situations like this that you can imagine occurring, with the differences in which HTTP/JSON/PSQL escape data?

EDIT: After trying some different things, it seems like this is definitely a HTTP-centered issue, I'll just document it.

KlaasH (approved):
:+1: Looks good! :tada:

PR #94, opened 2018-08-08
Feature/refactor function signatures

### LINKED ISSUE #86: Refactor tiler to slim down function signatures
This is a very low priority issue, but the tiler module contains a lot of repetitive function signatures and seems like it could be more DRY.

mattdelsordo:
## Overview

Because `image()`, `grid()`, and `vectorTile()` all share the same functional structure of 
1. Create a mapnik Map
2. Render the Map to the intended data format
3. Encode that data format to a form that can be transferred over the internet

I thought it made sense to refactor them such that the code format is consistent between functions. The main thing this PR does is change the function signatures in question to now accept a Promise<Map> + whatever other info the function may need to operate, rather than pass around 4+ variables that only get used by `createMap()`. I've also removed the promise-creating function that only `image()` used for encoding, and renamed the functions to share the same sort of naming scheme.

No actual functionality has been changed, I just thought it might be nice to streamline things a bit.

Resolves #86

### DIFF
--- a/src/tiler/src/api.js
+++ b/src/tiler/src/api.js
@@ -4,7 +4,7 @@
 
 import APIBuilder from 'claudia-api-builder'
 
-import { image, grid, vectorTile, createMap } from './tiler'
+import { imageTile, utfGrid, vectorTile, createMap } from './tiler'
 import HTTPError from './util/error-builder'
 
 const IMAGE_HEADERS = {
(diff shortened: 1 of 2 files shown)


KlaasH (approved):
A bit more terse and a bit more regular, looks good to me :+1:

PR #95, opened 2018-08-09
Refactor error handling to dynamically decide status codes

### LINKED ISSUE #75: Consider switching error handling to returning relevant HTTP response codes
Currently, endpoints that throw errors return the errors as a json or image message to the user in order to aid debugging and make it very visible that something went wrong. However, after some thought (and not being able to print error messages to vector tiles) I'm less confident in this design choice. My reasons for this are:

1. Messages can't be returned for all endpoints, which disrupts your expectation of how the tool in general is going to behave. Ideally, if there's an error, all endpoints should fail in the same, expected way.
2. Returning human-readable messages is great for human debuggers but might lead to unnecessary  debugging complications if a library used to fetch tiles has its own ways to gracefully handle HTTP status codes. 

It might be a good idea to revisit error handling in the api module and convert thrown errors into meaningful status codes, instead of returning the error content with a 200.

### LINKED ISSUE #76: Return an error to requests for nonexistent layers
The layer-filtering logic loops over the configured layers and turns off ones that aren't in the requested list.  But that means it just ignores entries in the list that don't match a configured layer.  Seems like returning an error if a layer is listed in the request but not defined in the map config would make sense, and failing loudly rather than silently could make it easier to debug typos or misconfiguration.

mattdelsordo:
## Overview

This PR removes the previous error handling system and replaces it with one that returns variable HTTP status codes depending on the situation. The goal was to be able to give users/tools that use the API more feedback about things going wrong, as well as intercept Postgis errors in order to prevent database info leaking out through error messages. Additionally, the removal of `jimp` as a dependency slims down the bundle size, which is a nice side effect. 

Error cases that I added:
 * Conversion of Postgis errors to simply "Postgis error" to prevent database info leakage
 * 404 response on all undefined endpoints
 * 400 status if there doesn't exist a layer for a given value of '&layers'
 * Misc error handlers for forseen common API typos

### Demo

image

### Notes
 * CloudFront will often not forward along HTTP statuses from API Gateway, returning a 500 'Internal server error' status rather than the actual, more specific error that API Gateway sends. This is annoying an I'm investigating it, but for the time being it isn't the end of the world since a) it still works a lot of the time and b) before, it wasn't even sending any error.
 * This PR should be merged ASAP so that other branches (specifically `feature/s3-config`) can take advantage of the reduced bundle size. Adding `aws-sdk` to the project in that branch made the project un-deployable, again. 
 * While not necessary for production, this PR could significantly benefit from changes I've made to mattdelsordo/claudia-local-api. The development server will crash unless either those changes are merged or you comment out the wildcard/404 endpoint that this PR adds.

## Testing Instructions

 * Start the development server with `./scripts/update` and `./scripts/server`
* You may want to comment out https://github.com/azavea/tilegarden/blob/f5a39163de911a7812216848074642cdbc179a9c/src/tiler/src/api.js#L175-L184 so that the development server doesn't crash (which means you can't test the 404 endpoint in development, unfortunately)
 * Visit [`localhost:3000/`](http://localhost:3000/) and enter some malformed URLs to see what happens

Resolves #75, resolves #76


src/tiler/src/api.js
```diff
@@ -135,5 +172,16 @@ api.get(
     HTML_RESPONSE,
 )
 
+// 404 response
+// This works in production but breaks the dev server just by existing
+api.get(
```
mmcfarland, line 177: I think we'll need a more accommodating solution if we intend to keep this in.  Having to un/comment out this on develop doesn't seem long term sustainable.

mattdelsordo: I've actually PR'd a fix for this on `claudia-local-api`, it just hasn't been bubbled up to Tilegarden yet.

src/tiler/src/util/layer-filter.js
```diff
@@ -19,6 +21,14 @@ const parsePromise = xmlString => new Promise((resolve, reject) => {
  * Set the status of each layer NOT in the list to false to disable
  */
 const filter = (xmlJson, enabledLayers) => {
+    const dne = enabledLayers.filter(a =>
+        xmlJson.Map.Layer.filter(b =>
+            b.$.name === a).length === 0)
```
KlaasH: The two closing parens on the last line are odd at first glance and a little tricky to track down. I wonder if removing the last line break (i.e. putting the whole expression and the closing paren of the the `a` filter function on one line) would be an improvement.

KlaasH (approved):
:+1:

src/tiler/src/util/layer-filter.js
mattdelsordo: That seems to play ball with ESLint at least (which is why I ended up formatting it like this). At any rate, this is completely replaced in the filtering PR.

PR #97, opened 2018-08-10
Ensure arguments get passed to the right utilities in the right order

mattdelsordo:
## Overview

On some machines the `--abbrev 300` argument in the `local` yarn script would get passed to node instead of to claudia-local-api. Added a `--` to make sure that the order of operations would be forced to work out.


mmcfarland (approved):
I encountered this problem and this patch resolves it +1

PR #101, opened 2018-08-15
Feature/demo performance

mattdelsordo:
## Overview
This PR makes some changes to improve setup speed in the demo, just to make the whole thing run more seamlessly for demonstration. Specifically it:
 * Updates the defaults in `env-template` to give a better idea of what performance requirements are normal.
 * Speed up CloudFront deployment time by using the lowest breadth price class by default.
 * Modify vector tile demo to request one layer at a time, for smaller individual tile sizes.
 * Increase zoom level on all demos for performance reasons.

### Notes
A TBD aspect of this pull request is updating the timeout/memory settings in the `.env` on the GitHub pages branch. I'm going to wait on doing that, though, until after the talks, because I very much do not want to break/spend time fixing the demo branch right now.

## Testing Instructions

 * Spin up a development server with `./scripts/update`/`./scripts/server`
 * Check out the updated demo pages, which should load faster overall.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,8 +88,8 @@ A deployed instance of Tilegarden consists of several AWS resources:
 
 #### Additional Configuration Options
  * `USER`: Left unspecified so that the code knows who the local user is from inside docker containers. Gets tacked on to the end of the project name at deployment.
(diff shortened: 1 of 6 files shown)


mmcfarland (approved):
+1, changes look good and demos are much speedier!

PR #102, opened 2018-08-17
Upgrade claudia-local-api to take advantage of local fixes

mattdelsordo:
## Overview
Upgrade `claudia-local-api` to take advantage of local fixes. Specifically, properly parsing API gateway-style wildcard URLs and not converting to binary if the status code indicates an error.


KlaasH (approved):
:+1:

PR #106, opened 2018-08-20
Feature/docker image

mattdelsordo:
## Overview

This PR adds the creation of a docker image from the tiler container which is then published along with the demo site on merges to develop. The `terraform` container has been removed from the development environment and it's functionality has been added to the `tiler` container, so that all deployment steps can be done from a single image. Main functionality for `build-xml`, `publish`, and `destroy` has also been moved into the `tiler` container to free users from requiring extra external scripts.

### Demo

Optional. Screenshots, `curl` examples, etc.

### Notes

Optional. Ancillary topics, caveats, alternative strategies that didn't work out, anything else.

## Testing Instructions

 * How to test this PR
 * Prefer bulleted description
 * Start after checking out this branch
 * Include any setup required, such as bundling scripts, restarting services, etc.
 * Include test case and expected output

Resolves #XXX

PR #108, opened 2018-08-21
Feature/streamline deployment

### LINKED ISSUE #46: Improve tool portability/ease of use
I want to be able to go from downloading the tool to having a functional "tileserver" live on AWS in as few button presses as possible.

mattdelsordo:
## Overview
This PR aims to streamline deployment by
1. giving `env-template` a name that better indicates what you're meant to do with it.
2. creating a `deploy` script that basically just calls `docker-compose build` before `./scripts/publish`

The state of this PR is sort of temporary as we work out what all should be included in it vs. #106 

## Testing Instructions

 * Run `./scripts/clean` on your development environment (or just make a fresh clone if you don't have the project on your machine)
 * Set up your project for deployment without running the docker containers, and then run `./scripts/deploy --new`. It should work where `./scripts/publish` wouldn't as it builds containers as well as publishing.

Resolves #46

### DIFF
--- a/scripts/deploy
+++ b/scripts/deploy
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+set -e
+
+deploy_new=""
+no_cache=""
+
+function usage() {
+    echo -n "Usage: $(basename "${0}") [OPTION]
+Deploys a Tilegarden instance to AWS by building docker containers first
+Options:
+    --help        Show usage
+    --new        Deploy as a new lambda
+    --no-cache    Don't use cache when building containers
+"
+    exit 0
+}
+
+function main() {
+	if [[ $deploy_new != "--new" && ! -f /src/tiler/claudia.json ]]; then
+		echo "No existing deployment found to update! (did you mean to use '--new'?)"
+		exit 1
+	fi
+
+    docker-compose build terraform
+
+	if [ "${no_cache}" == "--no-cache" ]; then
+    	docker-compose build --no-cache tiler
+    else
+    	docker-compose build tiler
+    fi
+
+	# Capture output to check for specific error
+	./scripts/publish "${deploy_new}"
+}
+
+
+if [ "${BASH_SOURCE[0]}" = "${0}" ]
+then
+    # handle command line arguments
+    # modified from https://stackoverflow.com/a/7069755
+    while [ $# -gt 0 ]; do
+        case "$1" in
+            -n| --new)
+                deploy_new='--new'
+                shift
+                ;;
+            --no-cache)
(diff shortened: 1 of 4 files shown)


scripts/deploy
```diff
@@ -0,0 +1,59 @@
...
+    	docker-compose build --no-cache tiler
+    else
+    	docker-compose build tiler
+    fi
+
+    if ./scripts/publish "${deploy_new}" ; then
+        :
+    else
+    	echo ""
+        echo "Did you maybe mean './scripts/deploy --new'?"
```
KlaasH: I cleared by `AWS_REGION` to see what would happen and ran `./scripts/deploy --new`.  It did this:
```
AWS region is missing. please specify with --region
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: sh
Arguments: -c yarn transpile && claudia create --api-module bin/api --name ${PROJECT_NAME} --region ${LAMBDA_REGION} ${LAMBDA_ROLE:+--role ${LAMBDA_ROLE}} ${LAMBDA_TIMEOUT:+--timeout ${LAMBDA_TIMEOUT}} ${LAMBDA_MEMORY:+--memory ${LAMBDA_MEMORY}} ${LAMBDA_SECURITY_GROUPS:+--security-group-ids ${LAMBDA_SECURITY_GROUPS}} ${LAMBDA_SUBNETS:+--subnet-ids ${LAMBDA_SUBNETS}} && yarn parse-id
Directory: /home/tiler
Output:
".
info If you think this is a bug, please open a bug report with the information provided in "/home/tiler/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Did you maybe mean './scripts/deploy --new'?
```
We probably shouldn't suggest this possibility for all errors.  At the least, we shouldn't suggest it when the `--new` option was in fact given.

KlaasH (changes requested):
I hit a variety of errors after the one mentioned below, mainly related to things being in partially deployed state.  That might be something to work on, but it's outside the scope of this PR.  So this looks good except for the "Did you maybe mean" message show up too often.

scripts/deploy
```diff
@@ -0,0 +1,60 @@
...
+
+    docker-compose build terraform
+
+	if [ "${no_cache}" == "--no-cache" ]; then
+    	docker-compose build --no-cache tiler
+    else
+    	docker-compose build tiler
+    fi
+
+	# Capture output to check for specific error
```
KlaasH, line 33: stale comment

KlaasH (approved):
:+1: Looks good!

PR #124, opened 2018-09-27
Use pre-installed AWS SDK on Lambda

### LINKED ISSUE #118: Shrink aws-sdk
Related to #117 

We only really use the fetching-from-S3 part of `aws-sdk`, so it might be feasible to remove the parts of it that aren't used in order to slim down the bundle size. Rumor has it that it's available locally in the Lambda Node runtime, so check if that is true. If so, it might be possible to just relegate `aws-sdk` to a `devDependency`.

KlaasH:
## Overview

The Lambda environment comes with the AWS SDK pre-installed, so it wastes space in the bundle to install it ourselves.

Also bumps the version listed in package.json (and installed in development) to the [current version pre-installed on Lambda](https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html)

### Notes
    
- Per [this issue](https://github.com/claudiajs/claudia/issues/82), just moving it to `devDependencies` won't work because Claudia validates the bundle and it's an error for the package to be missing, but that can be worked around by making at an optional dependency and telling Claudia not to install optional.
- This also adds a `--no-deps` argument to a few of the `docker-compose` commands in places where the `tiler` container is being used but the `database` container (on which it depends via the `links:` parameter) isn't needed.

## Testing Instructions

- To test local development, delete `node_modules` or run `scripts/clean` to clear the existing installation, then run `scripts/update` to build the tiler container again.  It should still install `aws-sdk` and work normally.
- To test the deployed package, run `scripts/publish` or `scripts/publish --new`.  Export the Lambda function to a zip file and confirm that `aws-sdk` is not present in the `node_modules` directory.
- If you already had a Lambda function deployed, you should be able to export the previous version and confirm that the size of the bundle has decreased.

Resolves #118.

### DIFF
--- a/scripts/clean
+++ b/scripts/clean
@@ -11,7 +11,7 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         usage
     else
         # Remove build artifacts from local machine
-        docker-compose run tiler rm -rf bin/ node_modules/
+        docker-compose run --no-deps tiler rm -rf bin/ node_modules/
 
         # Stop containers and shut down the network
         docker-compose down --volumes
--- a/scripts/destroy
+++ b/scripts/destroy
@@ -22,11 +22,11 @@ function main() {
         # If any part of this fails, the parts that didn't fail
         # remain destroyed, so it's best to just carry on and
         # follow through.
-        docker-compose run tiler yarn destroy || true
+        docker-compose run --no-deps tiler yarn destroy || true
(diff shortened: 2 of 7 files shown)


mmcfarland (approved):
Deploy package is about 2MB smaller :+1:

mtford90:
Is this live yet?

jeancochrane:
@mtford90 It is in the sense that it's been pulled into `develop`, which we use as the base branch for the repo. Currently we recommend installing and using the library by cloning the repo with Git, so you should be able to access these changes if you're using it that way.

